PR #16, opened 2024-04-17
helpful error messages

### LINKED ISSUE #15: register error hints?
I found [this package](https://github.com/jkrumbiegel/MakiePkgExtTest/tree/master) that illustrates makie recipes. In there, [this is shown:](https://github.com/jkrumbiegel/MakiePkgExtTest/blob/c1718640f96ba0dc72d2baaa489a1fb1097bfbc5/SomePackage/src/SomePackage.jl#L20C1-L34C4)
i.e.:
```julia
# in src/Visualize.jl
function __init__()
    Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
        if exc.f in [heatmap2D!, sliced_heatmap!, heatmap!, line_plot1D!, sliced_line_plot!, line_plot!, sliced_plot!, plot!, heatmap2D_on_globe!]
            if isempty(methods(exc.f))
                print(io, "\n$(exc.f) has no methods, yet. Makie has to be loaded for the plotting extension to be activated. Run `using Makie`, `using CairoMakie`, `using GLMakie` or any other package that also loads Makie.")
            end
        end
    end
end
```

As the readme states, it is intended:
`To guide unknowing users, I've implemented an error hint that only fires for MethodErrors on the plotting functions if they don't yet have any methods, and tells users that Makie needs to be loaded.`

With that, we would go from e.g.:
```julia
julia> ClimaAnalysis.Visualize.heatmap2D!(var2D)
ERROR: MethodError: no method matching heatmap2D!(::ClimaAnalysis.OutputVar{Vector{Float64}, Base.ReshapedArray{Float64, 2, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Tuple{}}, String, Dict{String, String}})
```
to
```julia
julia> ClimaAnalysis.Visualize.heatmap2D!(var2D)
ERROR: MethodError: no method matching heatmap2D!(::ClimaAnalysis.OutputVar{Vector{Float64}, Base.ReshapedArray{Float64, 2, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Tuple{}}, String, Dict{String, String}})
heatmap2D! has no methods, yet. Makie has to be loaded for the plotting extension to be activated. Run `using Makie`, `using CairoMakie`, `using GLMakie` or any other package that also loads Makie.
```

haakon-e:
Closes #15

## Content

----
- [ ] I have read and checked the items on the review checklist.

### DIFF
--- a/src/Visualize.jl
+++ b/src/Visualize.jl
@@ -18,4 +18,14 @@ function plot! end
 
 function heatmap2D_on_globe! end
 
+function __init__()
+    Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
+        if exc.f in [heatmap2D!, sliced_heatmap!, heatmap!, line_plot1D!, sliced_line_plot!, line_plot!, sliced_plot!, plot!, heatmap2D_on_globe!]
+            if isempty(methods(exc.f))
+                print(io, "\n$(exc.f) has no methods, yet. Makie has to be loaded for the plotting extension to be activated. Run `using Makie`, `using CairoMakie`, `using GLMakie` or any other package that also loads Makie.")
+            end
+        end
+    end
+end
+
 end


Sbozzolo:
Can you please format it?

Sbozzolo:
I don't think this really works as it is. For example, if I call `ClimaAnalysis.Visualize.plot!` instead of `plot!`, I'll get that
```
exc.f = ClimaAnalysis.Visualize.plot!
```
so the error hint is not triggered

Sbozzolo:
Closed by #184

PR #24, opened 2024-06-04
Add interpolation and interpolation to pressure levels

### LINKED ISSUE #23: Add support for interpolating to pressure levels
Ideally, more generally to new grids.

Sbozzolo:
This PR adds interpolation routines to OutputVar. With this PR, one can evaluate an `OutputVar` as if it were a mathematical function (`var(t, x, y, z)`) and the result is computed with multilinear interpolation.

I also added a method to interpolate onto pressure levels. At this moment, the target pressure levels are chosen by linearly sampling the interval from min to max.

When values are outside of the range, linear extrapolation is performed.

Closes #23


docs/src/howdoi.md
```diff
@@ -49,3 +49,11 @@ where inside `ca_kwargs` you pass the arguments you would pass to `Makie.Axis`.
 ## How do I center my longitude to 180 instead of 0?
 
 You can use the `center_longitude!` function.
+
+## How do I move to pressure coordinates?
+
+The `Atmos` module in `ClimaAtmos` comes with a function,
```
szy21, line 55:
```suggestion
The `Atmos` module in `ClimaAnalysis` comes with a function,
```

src/Atmos.jl
```diff
@@ -0,0 +1,109 @@
...
+    z_index = var.dim2index[z_name]
+    pressure_name = short_name(pressure)
+
+    # First, we construct the target pressure grid. For this, we take the
+    # extrema of pressure and divide the interval linearly with the same number
+    # of points we originally had in z
+
+    # TODO: Pick this more sensibly
+    # TODO: Make it go from max to min? (This is not supported by Interpolations.jl...)
+    target_pressure = range(
```
szy21, line 69: I think it would be more useful to fix the pressure levels or have it as an input argument for `to_pressure_coordinates`. For example, if we want to compare two different simulations (which will have different pressures), we want to compare the output on the same pressure levels.

szy21 (commented):
Thanks! Just a quick comment for now. I will take a closer look this week, but feel free to merge if you are in a hurry.

src/Atmos.jl
Sbozzolo: Added

NEWS.md
```diff
@@ -0,0 +1,11 @@
+ClimaAnalysis.jl Release Notes
+===============================
+
+v0.5.3
+------
+
+- Add support for evaluating `OutputVar`s onto arbitrary target points (with
+  multilinear interpolation).
+
+- Add support for evaluating `OutputVar`s onto arbitrary target points (with
```
akshaysridhar, line 10: duplicate info

src/Atmos.jl
```diff
@@ -0,0 +1,124 @@
...
+    var::OutputVar,
+    pressure::OutputVar;
+    target_pressure = nothing,
+)
+    arecompatible(var, pressure) ||
+        error("Pressure and variable are not compatible")
+
+    ALTITUDE_NAMES = Set(["z", "z_physical", "z_reference"])
+
+    # Pick the correct longitude name and check that we have an altitude variable
```
akshaysridhar, line 61:
```suggestion
    # Pick the correct altitude name and check that we have an altitude variable
```

akshaysridhar:
Minor comments in the docs / info blocks, otherwise this looks good. Thanks for adding tests as well.

src/Atmos.jl
Sbozzolo: Thank you!

Sbozzolo:
For some reason, codecov is not picking the coverage of the new file. I'll figure this out later.

szy21 (approved):
Thank you!

PR #28, opened 2024-06-13
Fix bug in window function

dennisYatunin:
## Purpose 

Followup to #18, which had a bug in the `window` function for multi-dimensional `OutputVar`s. Since `var.dims[dim_name]` is a `Vector`, it can only be sliced along its first dimension.

----
- [x] I have read and checked the items on the review checklist.


Sbozzolo (approved):

Sbozzolo:
CI is broken because I extended the compat for GeoMakie. I need to fix that

PR #42, opened 2024-07-25
Add ClimaAnalysisInteractive

Sbozzolo:
This commit adds an interactive viewer for ClimaAnalysis.

Work in progress!

image


AlexisRenchon:
I see for now ClimaExplorer is not using [`Observable`](https://docs.makie.org/v0.21/explanations/observables) 

Observables makes things a bit smoother because it only updates one object (e.g., a dot or a line) and does not recreate parent objects (a figure, an axis, etc.). Currently ClimaExplorer recreates everything each time, which can be slow and clunky (resizing everything each time for example). 

Note that sometime you do want to recreate the whole figure, in which case [SpecApi](https://docs.makie.org/stable/explanations/specapi) is more responsive. - but I think in our case we want Observables ideally

docs/src/explorer.md
```diff
@@ -0,0 +1,42 @@
...
+```sh
+julia --project=ClimaAnalysis.jl/climaexplorer -e 'using Pkg; Pkg.instantiate()'
+```
+
+How to use
+==========
+
+Once you installed `ClimaExplorer`, you can start the server with
+```sh
+julia --project=ClimaAnalysis.jl/climaexplorer ClimaAnalysis.jl/climaexplorer/explorer.jl PATH_OF_OUTPUT
```
AlexisRenchon, line 25: Maybe it would be nice to have an example output file inside of ClimaAnalysis that works out of the box? 

I wanted to test it, followed these docs, and realized I don't have an output path to test

climaexplorer/libexplorer.jl
```diff
@@ -0,0 +1,78 @@
...
+        end
+
+        var = map(level_slider) do level
+            short_name = vars_menu.value.val
+            ClimaAnalysis.slice(get(simdir; short_name, reduction = reduction.val, period = period.val); time = time_slider.value.val, z = level)
+        end
+
+        fig = map(var) do sliced_var
+            fig = CairoMakie.Figure(size=(1200, 760), fontsize=30)
+            ClimaAnalysis.Visualize.contour2D_on_globe!(fig, sliced_var)
```
AlexisRenchon, line 53: I think the way to go to make this a plot element (contour) `Observable` is to make distinct functions for the creation of the Axis (GeoAxis here), and the function `ClimaAnalysis.Visualize.contour2D_on_globe!` taking fig and axis as arguments, so that only the plot element (contour) gets updated dynamically - fig and axis remain the same, and don't need to be updated here (so fig should be out too, and defined only once)

ClimaExplorer/explorer.jl
```diff
@@ -0,0 +1,15 @@
+include(joinpath(@__DIR__, "ClimaExplorer.jl"))
+
+import Bonito: route!, wait, Server
+
+path = length(ARGS) >= 1 ? ARGS[1] : "."
+
+# Create server
+IPa = "127.0.0.1"
+port = 8080
+server = Server(IPa, port; proxy_url = "http://localhost:9384")
```
AlexisRenchon, line 10: Note: 
This is valid only for the remote cluster method. 
If run locally, you only need `server = Server(IPa, port)`
OR `server = Server(IPa, port; proxy_url = "http://localhost:8080")`

otherwise it will return a broken pipe error

docs/src/explorer.md
```diff
@@ -0,0 +1,61 @@
...
+```sh
+julia --project=ClimaAnalysis.jl/ClimaExplorer -e 'using Pkg; Pkg.instantiate()'
+```
+
+How to use
+==========
+
+Once you installed `ClimaExplorer`, you can start the server with
+```sh
+julia --project=ClimaAnalysis.jl/ClimaExplorer ClimaAnalysis.jl/ClimaExplorer/explorer.jl
```
AlexisRenchon, line 25: This will currently crash, see my comment above.

docs/src/explorer.md
```diff
@@ -0,0 +1,61 @@
...
+How to install
+===============
+
+To install `ClimaExplorer`, first clone the `ClimaAnalysis.jl` repository
+```sh
+git clone https://github.com/CliMA/ClimaAnalysis.jl.git
+```
+Then, instantiate the `climaexplorer` environment
+```sh
+julia --project=ClimaAnalysis.jl/ClimaExplorer -e 'using Pkg; Pkg.instantiate()'
```
AlexisRenchon, line 17: This currently needs to be `dev`, until we make a release

AlexisRenchon:
3140a7b switch to WGLMakie, but all I did is copy-paste the CairoMakieExt and replace with WGLMakie, likewise in scripts. 

This works, and the app is much faster and more responsive. 

However, you may want a different design (compatibility with both CairoMakie and WGLMakie? or no duplicate of code in Ext?) so feel free to update or delete this commit and rework it

ext/WGLMakieExt.jl
```diff
@@ -0,0 +1,549 @@
...
+    # be passed down and lead to a unknown argument error.
+    #
+    # TODO: Refactor: We shouldn't have to deal with dim_on_y if we don't use it!
+    if haskey(axis_kwargs, :dim_on_y)
+        axis_kwargs_dict = Dict(axis_kwargs)
+        pop!(axis_kwargs_dict, :dim_on_y)
+        axis_kwargs = pairs(axis_kwargs_dict)
+    end
+
+    WGLMakie.Axis(place[p_loc...]; title, xlabel, ylabel, axis_kwargs...)
```
AlexisRenchon, line 79: I think we will need to have separate function to create Axis and elements (like heatmap), in order to create the Axis only once outside of the interaction

Sbozzolo:
> [3140a7b](https://github.com/CliMA/ClimaAnalysis.jl/commit/3140a7b5dd883ca99ed1cc3464ef6261f0c611a3) and [c5d50d2](https://github.com/CliMA/ClimaAnalysis.jl/commit/c5d50d2f23746eb2a9b311a5225c26a3c1fe116f) switch to WGLMakie, but all I did is copy-paste the CairoMakieExt and replace with WGLMakie, likewise in scripts.
> 
> This works, and the app is much faster and more responsive.
> 
> However, you may want a different design (compatibility with both CairoMakie and WGLMakie? or no duplicate of code in Ext?) so feel free to update or delete this commit and rework it
> 
> (Note, I wanted to squash these 2 commits into one, and I did locally, but I am a bit worried that if I do git push -f it will remove previous commits, so for now I leave these 2)

I think there should be a way to have a single `MakieExt` that can use different backends, but I haven't looked at how to accomplish this

ClimaExplorer/explorer.jl
Sbozzolo: When run explorer.jl on my computer like this, it works

AlexisRenchon: For me it just loses interactivity (but initial figure shows)

PR #49, opened 2024-07-31
Add support for nanmean

### LINKED ISSUE #3: Compute mean ignoring NaNs
It would be good to add a function that computes the mean while ignoring NaNs.

Sbozzolo:
Closes #3


src/Var.jl
```diff
@@ -259,11 +260,12 @@ function average_lat(var; weighted = false)
 
             weights[index_tuple...] .= weights_1d
         end
-        weights ./= mean(weights_1d)
+        weights ./= nanmean(weights_1d)
```
szy21, line 263: For weighted nanmean, I think we should make the weights NaN when the data is NaN? Otherwise the average is not accurate as we include the weight of the NaN data (as cos(lat) would not be NaN) when calculating the weight.

Sbozzolo: Yes, right!

Sbozzolo: Fixed, do you want to have a second look?

PR #52, opened 2024-08-04
Do not specialize on CairoMakie

Sbozzolo:
Makie has multiple backends. Each backend loads the Makie package. Previously, ClimaAnalysis provided an extension for one particular backend, now, the extension was generalized and it should work for any backend.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ main
   multilinear interpolation).
 - `average` operations now ignore `NaN`s by default.
 - Add `has_*` methods to query whether a `Var` has a given dimension (e.g., `z`).
+- Support `Makie` backends besides `CairoMakie`.
 
 v0.5.6
 ------
--- a/Project.toml
+++ b/Project.toml
@@ -12,23 +12,24 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [weakdeps]
-CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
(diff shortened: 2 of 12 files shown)


AlexisRenchon:
Where do you specify which backend to use?

Sbozzolo:
> Where do you specify which backend to use?

As a user, you just load it. See the `CairoMakie` example. You can also use the `activate!` function 

https://docs.makie.org/v0.21/explanations/backends/backends

AlexisRenchon:
It's always been a bit weird to me. 
it works well if you import all backends:
```
julia> using CairoMakie

julia> using WGLMakie

julia> CairoMakie.activate!()

julia> WGLMakie.activate!()
```
which is what we do in ClimaLandSimulations
but if you just import Makie, how do you do it?
```
julia> using Makie

julia> CairoMakie.activate!()
ERROR: UndefVarError: `CairoMakie` not defined
Stacktrace:
 [1] top-level scope
   @ REPL[5]:1 
```

Sbozzolo:
I don't know the details, but I think this is how `Makie` implements everything in a backend-agnostic way.

The `Makie` package implements the logic and provides functions. Once you load a backend, the functions are complete and can be used.

You cannot simply load `Makie`, and every time you load a backend, `Makie` is also implicitely loaded. 

```julia
julia> using Makie
julia> using CairoMakie
julia> Makie.Figure()
.....
```

```julia
julia> using Makie
julia> using WGLMakie
julia> Makie.Figure()
.....
```

This allows us to write code that does not depend on the backend. Looking at other packages, I feel that this is the proper way to write a Makie extension. I should have done this since the beginning, but I only used `CairoMakie` so I never bothered checking if it worked as expected.

AlexisRenchon (approved):

Sbozzolo:
Forgot to update the NEWS.md!

PR #60, opened 2024-08-15
Add Utils.split_by_season

### LINKED ISSUE #59: Port `split_by_season`
The function `split_by_season` is currently in ClimaCoupler and should be moved to ClimaAnalysis.

ph-kev:
Closes #59 - The function split_by_season is in ClimaCoupler and should be moved to ClimaAnalysis. Also, the Dates package is added as a dependency because split_by_season operate on dates.

### DIFF
--- a/src/Utils.jl
+++ b/src/Utils.jl
@@ -1,7 +1,14 @@
 module Utils
 
 export match_nc_filename,
-    squeeze, nearest_index, kwargs, seconds_to_prettystr, warp_string
+    squeeze,
+    nearest_index,
+    kwargs,
+    seconds_to_prettystr,
+    warp_string,
+    split_by_season
+
+import Dates
 
 """
     match_nc_filename(filename::String)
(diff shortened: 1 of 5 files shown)


Sbozzolo:
I forgot that stdlib packages have to be excluded by the downgrade test
```
      - uses: julia-actions/julia-downgrade-compat@v1
        with:
          skip: Dates, Test, Statistics
```

src/Utils.jl
```diff
@@ -248,4 +255,45 @@ function warp_string(str::AbstractString; max_width = 42)
     return strip(lstrip(return_str, '\n'))
 end
 
+"""
+    split_by_season(dates::AbstractArray{<: Dates.DateTime})
+
+Take an array of dates and 4 split the dates by seasons into 4 vectors.
```
Sbozzolo, line 261:
```suggestion
Return four vectors with `dates` split by seasons.  
```

Sbozzolo (approved):
Thank you!

PR #64, opened 2024-08-16
Update docstring for warp_string and increase max

Sbozzolo:
42 is a little too short, especially when defaults are used

See here as an example with 42

image


ph-kev (approved):
LGTM!

PR #66, opened 2024-08-23
Add Var.resampled_as

### LINKED ISSUE #65: Use new interpolators to add `resampled_as`, a function that resamples an `OutputVar` to the same dimensions as another one
Currently, in ClimaCoupler.jl, `resample` is a function that resample 2D data from `src_lonlat` to `dest_lonlat`. This function should be moved to ClimaAnalysis.jl and can be made to work more generally over any number of dimensions instead of only lon and lat. Additionally, unit checking should be added when porting over to ClimaAnalysis.jl to prevent resampling when units do not match.

ph-kev:
closes #65 - This PR add `Var.resampled_as` which has the same functionality as `resample` in ClimaCoupler. Additionally, `Var._check_dims_consistent` is made to help with checking that the `dim` in two OutputVar are consistent. Lastly, Var.dim_type is added to find the type of a dimension given the name of the dimension.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -27,7 +27,8 @@ export OutputVar,
     long_name,
     units,
     dim_units,
-    range_dim
+    range_dim,
+    resampled_as
 
 """
     Representing an output variable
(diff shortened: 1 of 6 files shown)


src/Var.jl
```diff
@@ -581,6 +583,85 @@ function arecompatible(x::OutputVar, y::OutputVar)
     return (x.dims == y.dims) && (x.dim_attributes == y.dim_attributes)
 end
 
+"""
+    _check_dims_consistent(x::OutputVar, y::OutputVar)
+
+Check the number, type, and unit of dimensions in `x` and `y` are consistent.
```
Sbozzolo, line 589:
```suggestion
Check if the number, type, and unit of dimensions in `x` and `y` are consistent.
```

src/Var.jl
```diff
@@ -581,6 +583,85 @@ function arecompatible(x::OutputVar, y::OutputVar)
...
+    (any(missing_x) && any(missing_y)) && error(
+        "Units for dimensions $(x_dims[missing_x]) are missing in x and units for dimensions $(y_dims[missing_y]) are missing in y",
+    )
+    any(missing_x) &&
+        error("Units for dimensions $(x_dims[missing_x]) are missing in x")
+    any(missing_y) &&
+        error("Units for dimensions $(x_dims[missing_y]) are missing in y")
+
+    # Check whether units match between dimensions 
+    not_consistent_units = .!(x_units .== y_units)
```
Sbozzolo, line 626:
```suggestion
    not_consistent_units = (x_units .!= y_units)
```

src/Var.jl
```diff
@@ -581,6 +583,85 @@ function arecompatible(x::OutputVar, y::OutputVar)
...
+    any(mismatch_dim_type) && error(
+        "Type of dimensions do not agree between x ($dim_type_x) and y ($dim_type_y)",
+    )
+
+    x_dims = collect(keys(x.dims))
+    y_dims = collect(keys(y.dims))
+    x_units = [dim_units(x, dim_name) for dim_name in x_dims]
+    y_units = [dim_units(y, dim_name) for dim_name in y_dims]
+
+    # Check for any missing units
```
Sbozzolo, line 614:
```suggestion
    # Check for any missing units (missing units are represented with an empty string)
```

src/Var.jl
```diff
@@ -581,6 +583,85 @@ function arecompatible(x::OutputVar, y::OutputVar)
...
+
+The resampling performed here is a 1st-order linear resampling. 
+"""
+function resampled_as(src_var::OutputVar, dest_var::OutputVar)
+    _check_dims_consistent(src_var, dest_var)
+
+    src_resampled_data = [
+        src_var.interpolant(pt...) for
+        pt in Iterators.product(collect(values(dest_var.dims))...)
+    ]
```
Sbozzolo, line 646:
```suggestion
    src_resampled_data = [
        src_var(pt...) for
        pt in Iterators.product(collect(values(dest_var.dims))...)
    ]
```

Is `collect` needed?

src/Var.jl
```diff
@@ -581,6 +583,85 @@ function arecompatible(x::OutputVar, y::OutputVar)
...
+
+    # Construct new OutputVar to return 
+    src_var_ret_dims = empty(src_var.dims)
+
+    # Loop because names could be different in src_var compared to dest_var
```
Sbozzolo, line 651:
```suggestion
    # Loop because names could be different in src_var compared to dest_var (e.g., `long` in one and `lon`) in the other
```

src/outvar_dimensions.jl
```diff
@@ -167,3 +168,18 @@ altitude_name(var::OutputVar) = find_dim_name(keys(var.dims), ALTITUDE_NAMES)
...
+if possible or `dim_name` as a symbol otherwise. 
+"""
+function dim_type(dim_name::AbstractString)
+    dim_name in LONGITUDE_NAMES && return :longitude
+    dim_name in LATITUDE_NAMES && return :latitude
+    dim_name in TIME_NAMES && return :time
+    dim_name in DATE_NAMES && return :date
+    dim_name in ALTITUDE_NAMES && return :altitude
+    return Symbol(dim_name)
+end
```
Sbozzolo, line 185: I suggest we rename this to `conventional_dim_name` because it is clearer than an undefined notion of "type". 

In that case, I am not sure on the return type, symbol or string.

Sbozzolo (commented):
Can you also add a FAQ and a small tutorial/example in the docs that showcases this feature?

Sbozzolo:
You should also add `Iterators` to the package dependencies (with compat `1`)

src/Var.jl
```diff
@@ -3,6 +3,7 @@ module Var
 import NCDatasets
 import OrderedCollections: OrderedDict
 
+import Dates
```
Sbozzolo, line 6: Is this needed?

docs/src/faq.md
Sbozzolo, line 1: I feel "FAQ" and "How do I?" are largely overlapping. What I had in mind was a more general "OutputVar" page. We can make work on that in a future PR, so maybe let's move this question to "How do I?" for the time being.

Sbozzolo (commented):
Please, add a NEWS entry under as a new "feature", showcasing how to use it (you can just copy and paste your FAQ).

Sbozzolo (approved):

Sbozzolo (approved):

PR #70, opened 2024-08-29
Add support for unitful

### LINKED ISSUE #53: Add support for `Unitful`
[Unitful.jl](https://github.com/PainterQubits/Unitful.jl) is a package that handles physical units. It would be interesting to explore a possible integration with `ClimaAnalysis`.

Sbozzolo:
This commit adds support for physical units managed by Unitful.jl.

Unitful is added as an extension. Currently, ClimaAnalysis.jl always includes this extension (and Unitful is a hard dependency of
ClimaAnalysis), but this opens a pathway to making Unitful an optional dependency in the future if we decide to do so. (I am taking this idea from Unitful itself.)

The technical choice on how to integrate Unitful is the following:

In the constructor, ClimaAnalysis scans for the `units` keyword in the attributes. When found, we try to parse it with 
`Unitful.uparse`. If possible, we replace the value for `units` in `attributes` with the `Unitful` version. If it is not possible, we leave the string.

When units are requested with `ClimaAnalysis.units`, we always return a string.

The choice being made here is that we modify the input given (from `String` to `Unitful`), instead of creating a separate box where to put this information. The reason for this choice is that it removes the need to keep information synced and provide a single ground truth of what units are.

We also decide to not attach units directly to `data` to avoid performance penalties with manipulating large arrays.

This PR also cleans up a couple of things about extensions along the way.

Closes #53 (except for dimensions)


AlexisRenchon (approved):
This a beautiful PR! 
I have one question (curiosity) but no request

Why did you define convert_units docstring in Var.jl, and the actual function in ext/ClimaAnalysisUnitfulExt.jl? 

Is it to keep ClimaAnalysisUnitfulExt focused on actual code, for better readability?

Sbozzolo:
> This a beautiful PR! I have one question (curiosity) but no request
> 
> Why did you define convert_units docstring in Var.jl, and the actual function in ext/ClimaAnalysisUnitfulExt.jl?
> 
> Is it to keep ClimaAnalysisUnitfulExt focused on actual code, for better readability?

I had it there originally, but Documenter wouldn't find the docstring.

PR #72, opened 2024-08-31
Add support for overplotting a ocean/land mask with `GeoMakie`

Sbozzolo:
This commit adds support for applying land/sea masks to GeoMakie plots. The masks are obtained directly from GeoMakie as a collection of polygons. For the ocean mask, we use bathymetry data at 0 elevation. Data is provided by
`NaturalEarth.jl`. (I had to bump the minimum compat for GeoMakie to 0.7 because support `NaturalEarth.jl` was introduced in that release.)

This new option is only aesthetics: it does change the underlying data (we simply overplot something else). This also means that the colorbar will include values under the mask. Applying the mask directly to the data is left for future work.

Thanks @haakon-e for help with this feature.

image


AlexisRenchon:
oooh that's cool!!

AlexisRenchon (approved):

AlexisRenchon:
"Applying the mask directly to the data is left for future work"

we could make it a clear outlier and use `lowclip` or `highclip` in the heatmaps, then build the colorbar without outlier (by building it independently), e.g.

```
ax, hm = heatmap(fig[1, 1], xs, ys, zs, colormap = :something,
    colorrange = (-x, y), lowclip = :blue)
Colorbar(fig[1, 4], limits = (-x, y), colormap=:something)
```

https://docs.makie.org/stable/reference/blocks/colorbar#colorbar

Sbozzolo:
> "Applying the mask directly to the data is left for future work"
> 
> we could make it a clear outlier and use `lowclip` or `highclip` in the heatmaps, then build the colorbar without outlier (by building it independently), e.g.
> 
> ```
> ax, hm = heatmap(fig[1, 1], xs, ys, zs, colormap = :something,
>     colorrange = (-x, y), lowclip = :blue)
> Colorbar(fig[1, 4], limits = (-x, y), colormap=:something)
> ```
> 
> https://docs.makie.org/stable/reference/blocks/colorbar#colorbar

My plan is to create an `OutputVar` from a land/sea mask so that it can be used beyond plotting.

PR #76, opened 2024-09-05
Add Var.integrate* functions

### LINKED ISSUE #75: Port `integrate_*` functions to allow for integrating of an `OutputVar`. In this, we follow the same pattern as the `average_*` functions.
Currently, ClimaCoupler.jl implements `integrate_on_sphere` to integrate the data on a sphere. The data must be a two dimensional array. This function should be ported over to ClimaAnalysis and extended to create three different functions `integrate_lonlat`, `integrate_lon`, and `integrate_lat`. These functions should also work if there is more than two dimensions in data for `integrate_lonlat` and more than one dimension in data for `integrate_lon` and `integrate_lat`.

ph-kev:
Closes #75 - This PR port the functionality of integrating `OutputVar` to ClimaAnalysis from ClimaCoupler. In particular, `Var.integrate_lonlat`, `Var.integrate_lon`, and `Var.integrate_lat` for integrating `OutputVar` with respect to longitude, latitude, or
both. Additionally, this commit adds a Numerics module to decouple the numerical implementation of integration and the construction of an `OutputVar`. The numerical implementation is rectangular integration using the midpoint rule when the points are equispaced which assume each point correspond to a midpoint of a cell and using the left endpoint rule when the points are not equispaced.

Lastly, `update_long_name_average!` is replaced by `update_long_name_generic!` so that the integrating functions can also use this function. Also, `_reduce_over` take in `args` and `kwargs` in `squeeze`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -120,6 +120,50 @@ contour2D_on_globe!(fig,
 CairoMakie.save("myfigure.pdf", fig)
 ```
 
+## Integrating `OutputVar` with respect to longitude or latitude
+
+You can use the `integrate_lon(var)`, `integrate_lat(var)`, or `integrate_lonlat(var)`
+functions for integrating along longitude, latitude, or both respectively. The bounds of
+integration are determined by the range of the dimensions longitude and latitude in `var`.
+The unit of both longitude and latitude should be degree.
+
+If the points are equispaced, it is assumed that each point correspond to the midpoint of a
+cell which results in rectangular integration using the midpoint rule. Otherwise, the
+integration being done is rectangular integration using the left endpoints for integrating
+longitude and latitude. See the example of integrating over a sphere where the data is all
+ones to find the surface area of a sphere.
+
+```@julia integrate_lonlat
+julia> lon = collect(range(-179.5, 179.5, 360));
+
+julia> lat = collect(range(-89.5, 89.5, 180));
+
+julia> data = ones(length(lon), length(lat));
+
+julia> dims = OrderedDict(["lon" => lon, "lat" => lat]);
+
+julia> dim_attribs = OrderedDict([
+           "lon" => Dict("units" => "degrees_east"),
+           "lat" => Dict("units" => "degrees_north"),
+       ]);
+
+julia> attribs = Dict("long_name" => "f");
+
+julia> var = ClimaAnalysis.OutputVar(attribs, dims, dim_attribs, data);
+
+julia> integrated_var = integrate_lonlat(var);
+
+julia> integrated_var.dims # no dimensions since longitude and latitude are integrated out
+OrderedDict{String, Vector{Float64}}()
+
+julia> integrated_var.data # approximately 4π (the surface area of a sphere)
+0-dimensional Array{Float64, 0}:
+12.566530113084296
+
+julia> long_name(integrated_var) # updated long name to reflect the data being integrated
+"f integrated over lon (-179.5 to 179.5degrees_east) and integrated over lat (-89.5 to 89.5degrees_north)"
+```
+
 ## Bug fixes
 
 - Increased the default value for `warp_string` to 72.
--- a/src/ClimaAnalysis.jl
+++ b/src/ClimaAnalysis.jl
@@ -4,6 +4,8 @@ import Reexport: @reexport
 include("Utils.jl")
 import .Utils
 
+include("Numerics.jl")
+
 include("Var.jl")
 @reexport using .Var
 include("Sim.jl")
--- a/src/Numerics.jl
+++ b/src/Numerics.jl
@@ -1,0 +1,128 @@
+module Numerics
+
+import ..Utils: _isequispaced
+
+"""
+    _integrate_lon(data::AbstractArray, lon::AbstractVector; dims)
+
+Integrate out longitude from `data`. `data` has to be discretized on `lon`.
+
+`dims` indicates which axis of `data` is longitude.
+
+If the points are equispaced, it is assumed that each point correspond to the midpoint of a
+cell which results in rectangular integration using the midpoint rule. Otherwise, the
+integration being done is rectangular integration using the left endpoints. The unit for
+longitude should be degrees.
+"""
+function _integrate_lon(data::AbstractArray, lon::AbstractVector; dims)
+    length(lon) == 1 &&
+        error("Cannot integrate when longitude is a single point")
+    _isequispaced(lon) ?
+    int_weights = _integration_weights_lon_equispaced(lon) :
+    int_weights = _integration_weights_lon_left(lon)
+    return _integrate_over_angle(data, lon, dims, int_weights)
+end
+
+"""
+    _integrate_lat(data::AbstractArray, lat::AbstractVector; dims)
+
+Integrate out latitude from `data`. `data` has to be discretized on `lat`.
+
+`dims` indicates which axis of `data` is latitude.
+
+If the points are equispaced, it is assumed that each point correspond to the midpoint of a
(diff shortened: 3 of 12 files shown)


src/Var.jl
```diff
@@ -500,21 +505,27 @@ function range_dim(var::OutputVar, dim_name)
...
+    _update_long_name_generic!(
+    reduced_var::OutputVar,
+    var::OutputVar,
+    dim_name,
+    operation_name,
+)
 
-Used by averaging functions to update the long name of `reduced_var` by describing what the
-average is being taken over and the associated units.
+Used by reducing functions to update the long name of `reduced_var` by describing the
```
Sbozzolo, line 519:
```suggestion
Used by reductions (e.g., average) to update the long name of `reduced_var` by describing the
```

NEWS.md
```diff
@@ -120,6 +120,46 @@ contour2D_on_globe!(fig,
 CairoMakie.save("myfigure.pdf", fig)
 ```
 
+## Integrating `OutputVar` with respect to longitude or latitude
+
+You can use the `integrate_lon(var)`, `integrate_lat(var)`, or `integrate_lonlat(var)`
+functions for integrating along longitude, latitude, or both respectively. The bounds of
+integration are determined by the range of the dimensions longitude and latitude in `var`.
+The convention is that latitude ranges from -90° to 90° and longitude ranges from -180° to
+180° and the unit of both longitude and latitude should be degree. See the example of
```
Sbozzolo, line 129: I don't think we need lon/lat to be in these ranges (it would still work with longs from 0 to 360): all we need is that they are in degrees.

docs/src/howdoi.md
```diff
@@ -107,3 +107,43 @@ OrderedDict{String, Vector{Float64}} with 2 entries:
   "lon"      => [0.0, 1.0]
   "latitude" => [0.0, 1.0, 2.0]
 ```
+
+## How do I integrate the data in a `OutputVar` with respect to longitude or latitude?
```
Sbozzolo, line 111: Let's move this block to the new `vars.md`, since this section is getting croweded. There, you can spin it as a feature instead of a question (with this I mean, just change the title and the wording a little bit)

The block should also mention how we perform integration (ie, with the rectangle rule)

src/ClimaAnalysis.jl
```diff
@@ -4,6 +4,9 @@ import Reexport: @reexport
 include("Utils.jl")
 import .Utils
 
+include("Numerics.jl")
+@reexport using .Numerics
```
Sbozzolo, line 8: If all your functions are private, you don't need reexporting

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
...
+    _integrate_lon(data::AbstractArray, lon::AbstractVector; dims)
+
+Integrate out longitude from `data`. `data` has to be discretized on `lon`.
+
+`dims` indicates which axis of `data` is longitude.
+
+If the points are equispaced, it is assumed that each point correspond to the midpoint of a
+cell which results in rectangular integration using the midpoint rule. Otherwise, the
+integration being done is rectangular integration using the left endpoints. The unit for
+longitude should be degrees. The convention is that longitude ranges from -180° and 180°.
```
Sbozzolo, line 22: The important convention here is that units are degrees. The range doesn't have to be between -180 and 180

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
+module Numerics
+
+"""
+    isequispaced(arr::Vector)
+
+Return whether the array is equispaced or not.
+"""
+function _isequispaced(arr::Vector)
```
Sbozzolo, line 8: This feels like it should live in Utils, not in Numerics.

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
+module Numerics
+
+"""
+    isequispaced(arr::Vector)
+
+Return whether the array is equispaced or not.
```
Sbozzolo, line 6: Maybe let's also add a couple of jldoctests

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
...
+end
+
+"""
+    _integration_weights_lon_left(lon::AbstractVector)
+
+Return integration weights for rectangular integration using left endpoints for integrating
+along longitude.
+"""
+function _integration_weights_lon_left(lon::AbstractVector)
+    d_lon = deg2rad.(diff(lon))
```
Sbozzolo, line 88:
```suggestion
   # This is where we use the assumption that units are degrees
      d_lon = deg2rad.(diff(lon))
```

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
...
+    return int_on_angle
+end
+
+"""
+    _integration_weights_lon_left(lon::AbstractVector)
+
+Return integration weights for rectangular integration using left endpoints for integrating
+along longitude.
+"""
+function _integration_weights_lon_left(lon::AbstractVector)
```
Sbozzolo, line 87: You probably don't need to specify the type in `lon` here

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
...
+end
+
+"""
+    _integration_weights_lat_left(lat::AbstractVector)
+
+Return integration weights for rectangular integration using left endpoints for integrating
+along latitude.
+"""
+function _integration_weights_lat_left(lat::AbstractVector)
+    d_lat = deg2rad.(diff(lat))
```
Sbozzolo, line 102: Same as above

src/Numerics.jl
Sbozzolo: Let's also mention that we check `length(lon) > 1` elsewhere

src/Numerics.jl
```diff
@@ -0,0 +1,132 @@
...
+end
+
+"""
+    _integration_weights_lon_equispaced(lon::AbstractVector)
+
+Return integration weights for rectangular integration when the points are equispaced for
+integrating along longitude.
+"""
+function _integration_weights_lon_equispaced(lon::AbstractVector)
+    return deg2rad.([lon[begin + 1] - lon[begin] for _ in lon])
```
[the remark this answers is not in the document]
Sbozzolo: But we can keep it like this for symmetry.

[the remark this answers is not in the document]
Sbozzolo: And this is also where we use the assumption we are in degress

src/Var.jl
```diff
@@ -785,6 +789,105 @@ function resampled_as(src_var::OutputVar, dest_var::OutputVar)
...
+        "deg",
+        "degs",
+        "°",
+        "degrees_north",
+        "degrees_east",
+    ]
+    angle_dim_unit = dim_units(var, angle_dim_name)
+    angle_dim_unit in deg_unit_names ||
+        lowercase(angle_dim_unit) |> (x -> occursin("deg", x)) ||
+        error("The unit for $angle_dim_name is missing or is not degree")
```
Sbozzolo, line 873: Can you just substitute this with
```
lowercase(angle_dim_unit) in deg_unit_names
```
?

test/runtests.jl
```diff
@@ -2,15 +2,16 @@ using SafeTestsets
 using Test
 
 #! format: off
-# @safetestset "Aqua" begin @time include("aqua.jl") end
-# @safetestset "Docstrings" begin @time include("doctest.jl") end
-# @safetestset "Format" begin @time include("format.jl") end
```
Sbozzolo, line 7: Thank you! I accidentally commented the tests out. Would you be able to split this change off to a different commit?

Sbozzolo (approved):
Awesome! Thank you!

Issue #82, opened 2024-09-09
Multiplication and division between two OutputVars does matrix multiplication and division


ph-kev:
Multiplying and dividing two OutputVars should do element wise multiplication and division instead. 

This issue can hopefully be resolved by adding `@.` to the lines doing multiplication and division between the data of the two OutputVars. Another solution is to support broadcasting so that `var1 .* var2` does element wise multiplication.

PR #83, opened 2024-09-10
Fix bug and change multiplication and division to be elementwise for OutputVars

### LINKED ISSUE #81: Bug: Interpolation does not update when doing +,-,*, or / between a Real and OutputVar
When doing something like `var + 1.0`, the interpolation does not update. This is because we are adding `1.0` to `var.data` in-place.

A solution to to this issue is to make a new OutputVar and return that instead.

ph-kev:
Closes #81, #82 - This PR fix a bug where interpolation do not update when a binary operation is performed between an `OutputVar` and a Real and change multiplication and division between `OutputVar`s to be elementwise multiplication and division instead of matrix multiplication and division.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -1007,36 +1007,52 @@ macro overload_binary_op(op)
             )
         end
         function Base.$op(x::OutputVar, y::Real)
-            ret_var = copy(x)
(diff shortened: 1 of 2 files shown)


Sbozzolo (approved):
Thank you!

PR #87, opened 2024-09-13
Add Visualize.plot_bias!

### LINKED ISSUE #86: Port `plot_bias`
This function lives in ClimaCoupler and should be ported to ClimaAnalysis.

ph-kev:
closes #86 - This PR adds `Visualize.plot_bias!` which is ported over from ClimaCoupler.  

See the plot below for what `Visualize.plot_bias!` produce. 
bias_plot


NEWS.md
```diff
@@ -253,6 +253,27 @@ julia> units(se_var)
 "K^2"
 ```
 
+### Plotting bias
+
+To plot the bias and display information such as the root mean squared error (RMSE) and the
+global bias in the plot, you use the function `plot_bias!(fig, sim, obs)`. In the example
+below, we plot the bias between our simulation and some observations stored in
+"ta\_1d\_average.nc".
```
Sbozzolo, line 261:
```suggestion
Building upon the other features introduced in this release, you can now directly plot bias and root
mean squared error between two variables with the `plot_bias!` function. Typically, this is done to 
compare simulated data against observations. 

In the example below, we plot the bias between our simulation and some observations stored in
`ta_1d_average.nc`.
```

src/Visualize.jl
```diff
@@ -28,4 +28,6 @@ function contour2D_on_globe! end
 
 function heatmap2D_on_globe! end
 
+function plot_bias! end
```
Sbozzolo, line 31:
```suggestion
function plot_bias_on_globe! end
```
I think we should be consistent with the other names in this function

Sbozzolo (approved):

Sbozzolo (approved):

PR #96, opened 2024-09-26
Add Var.reordered_as

### LINKED ISSUE #94: Add `reordered_as`, a function that reorders an `OutputVar` to match the ordering of dimensions of another one
As of now, `resampled_as` fails whether the ordering of dimensions do not match. If this is the case, then the dimensions should be reordered so that resampling is possible.

ph-kev:
closes #94 - This PR adds the function `reordered_as` which reorder the dimensions in a OutputVar to match the ordering of dimensions in another OutputVar. Furthermore, `resampled_as` is updated to use `reordered_as`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,36 @@
 ClimaAnalysis.jl Release Notes
 ===============================
 
+v0.5.9
+------
+
+## Features
+
+### Reorder dimensions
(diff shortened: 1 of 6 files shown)


Sbozzolo (approved):
Thank you!

PR #103, opened 2024-09-27
Fix empty units and bump to 0.5.9

Sbozzolo:
When units are "", Unitful doesn't know what to and fails. It doesn't
fail normally, instead, it fails with a `MethodError`. This commit
treats that case specially.


ph-kev (approved):
👍

PR #108, opened 2024-10-03
Add automatic conversion between dates in time dimension to seconds in time dimension and preprocessing for time dimension

### LINKED ISSUE #98: Add support for converting between dates to seconds for time/date dimension
When the time dimension is of type Dates.DateTime or there is a date dimension, we can't use functions like `resampled_as` which relies on the interpolant. One approach is to provide functions for converting dates to time. The function `dates_to_seconds(var::OutputVar; shift_to_end_of_month = false)` can be added which add a start date to `dim_attributes` using the first element of the time/date dimension and convert the dates to seconds with reference to the start date. The other function can be `choose_start_date(var::OutputVar, start_date::String)` which add or subtract an offset to adjust for the new start date.

### LINKED ISSUE #107: Var.arecompatiable is too strict
Var.arecompatible is too strict and should be relaxed to just check if the units are the same across the dimensions and not if the dimension attributes are the same between the two `OutputVar`s

ph-kev:
closes #98, closes #107 

This PR adds functions for handling conversions from dates in the time dimension to seconds in the time dimension, shifting the dates to the first day of the month, and relax the conditions for `Var.arecompatible`.

Conversion from dates to seconds in the time dimension is handled automatically at the constructor level by the function `Var._dates_to_seconds`. Shifting to the first day of the month and adding any offset in multiple of months is done by `Var.shift_to_beginning_of_month`. Finally, the condtions for `Var.arecompatible are relaxed to just check for the units between dimensions instead of checking that the dimension attributes match exactly.


NEWS.md
```diff
@@ -20,11 +20,44 @@ arrays are equispaced and span the entire range, then a periodic boundary condit
...
+preprocessing of dates is needed before shifting to `new_start_date`, then the parameter
+`shift_by` can be used as it accepts a function that takes in `Dates.DateTime` elements and
+return Dates.DateTime elements. This function is applied to each element of the time array.
+```julia
+# Shift the dates to first day of month, convert to seconds, and adjust seconds to
+# match "1/1/2010"
+shift_var = OutputVar(
+        "test.nc",
+        "pr",
+        new_start_date = "1/1/2010",
```
Sbozzolo, line 37: What formats are allowed?

Can it be a DateTime?

NEWS.md
```diff
@@ -20,11 +20,44 @@ arrays are equispaced and span the entire range, then a periodic boundary condit
 for the longitude dimension and a flat boundary condition is added for the latitude
 dimension.
 
+### Preprocess dates and times
+There is now support for preprocessing dates and times. The constructor for reading NetCDF
```
Sbozzolo, line 24: Can you describe where this function would be useful? (To adjust monthly averaged data)

docs/src/var.md
```diff
@@ -80,6 +80,41 @@ and latitude dimensions, a periodic boundary condition and a flat boundary condi
 added, respectively, when the dimension array is equispaced and spans the entire range. For
 all other cases, extrapolating beyond the domain of the dimension will throw an error.
 
+## Preprocess dates and seconds
+When loading a NetCDF file, dates in the time dimension is automatically converted to
```
Sbozzolo, line 84: Let's mention why we are doing this

docs/src/var.md
```diff
@@ -80,6 +80,41 @@ and latitude dimensions, a periodic boundary condition and a flat boundary condi
...
+seconds and a start date is added to the attributes of the `OutputVar`. Two additional
+parameters are provided to help preprocess dates which are `new_start_date` and `shift_by`.
+If `new_start_date` is provided, then dates in the time dimension will automatically be
+converted with reference to the `new_start_date` rather than the first date found in the
+NetCDF file. The parameter `new_start_date` can be any string parseable by the
+[Dates](https://docs.julialang.org/en/v1/stdlib/Dates/) module. If additional preprocessing
```
Sbozzolo, line 90: Can we make it compatible with passing a DateTime too?

src/Var.jl
```diff
@@ -1299,6 +1336,116 @@ function global_rmse(sim::OutputVar, obs::OutputVar)
...
+    start_date = time_arr[begin]
+    time_arr = map(date -> date_to_time(start_date, date), time_arr)
+
+    # Add offset to change to new start date if needed
+    if !isnothing(new_start_date)
+        new_start_date = Dates.DateTime(new_start_date)
+        dt = period_to_seconds_float(new_start_date - start_date)
+        time_arr .= time_arr .- dt
+        start_date = new_start_date
+    end
```
Sbozzolo, line 1384: Wouldn't this work?
```suggestion
    # Convert from dates to seconds using the first date in the time dimension array as the
    # start date or the new_start_date
    start_date = isnothing(new_start_date) ? time_arr[begin] : new_start_date
    time_arr = map(date -> date_to_time(start_date, date), time_arr)
```

src/Var.jl
```diff
@@ -1299,6 +1336,116 @@ function global_rmse(sim::OutputVar, obs::OutputVar)
...
+        conventional_dim_name(dim_name) == "time" ? dim_name => time_arr :
+        dim_name => dim_data for (dim_name, dim_data) in var_dims
+    )
+    ret_dims = OrderedDict(ret_dims_generator...)
+    ret_data = copy(var.data)
+    return OutputVar(ret_attribs, ret_dims, ret_dim_attribs, ret_data)
+end
+
+"""
+    shift_to_beginning_of_month(var::OutputVar; shift_by_months = 0)
```
Sbozzolo, line 1402: I think this function with the `shift_by_months` argument is a little confusing.

Maybe we can could have `shift_to_end_of_previous_month(var::OutputVar)` instead? This would ensure that the obs and sim are defined on the same date and that the seasons are correct.

Let's also add a use case to the description of the function

src/Var.jl
ph-kev: That works

src/Var.jl
ph-kev: Shouldn't this be called `shift_to_start_of_previous_month`?

docs/src/var.md
ph-kev: I changed it to allow DateTime now

src/Var.jl
Sbozzolo: Yes, right

Sbozzolo (approved):
Thank you!

PR #117, opened 2024-10-10
GeoMakie ext: no shading by default

AlexisRenchon:
GeoMakie extension `heatmap2D_on_globe!()` did not specify the default shading, therefore it had shading (Makie default), which made plots where obscured in some regions, making it difficult to clearly visualize colors. This commit fixes this issue by changing the default shading to noshading.

Before:
myfigure_default

After:
myfigure
 
Note that in some instances the shading was only at certain locations, making it even more confusing:
Screenshot 2024-10-10 at 12 18 10 PM

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,6 +52,7 @@ pfull_var = ClimaAnalysis.get(simdir, short_name = "pfull", reduction = "2.0d", 
 - `Var.shift_to_start_of_previous_month` now checks for duplicate dates and throws an error
 if duplicate dates are detected.
 - Fix issue with plotting multiple figures at the same time.
+- Improve shading for `Visualize.heatmap2D_on_globe!`.
 
 v0.5.10
 -------
--- a/ext/ClimaAnalysisGeoMakieExt.jl
+++ b/ext/ClimaAnalysisGeoMakieExt.jl
@@ -357,14 +357,8 @@ function Visualize.plot_bias_on_globe!(
         ),
         :cb => Dict(:ticks => ticks),
     )
-
-    # Function for recursively merging two dictionaries if the values of the dictionaries
-    # are dictionaries and the values of those are also dictionaries and so on
-    # See: https://discourse.julialang.org/t/multi-layer-dict-merge/27261/6
-    recursive_merge(x::AbstractDict...) = merge(recursive_merge, x...)
-    recursive_merge(x...) = x[end]
-    default_and_more_kwargs = recursive_merge(default_kwargs, more_kwargs)
-
+    default_and_more_kwargs =
+        ClimaAnalysis.Utils._recursive_merge(default_kwargs, more_kwargs)
     return Visualize.contour2D_on_globe!(
         place,
         bias_var;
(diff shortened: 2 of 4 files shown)


Sbozzolo (commented):
Is this relevant only for GeoMakie, or for Makie plots too?

AlexisRenchon:
this MWE 

```
using GeoMakie
using CairoMakie

fig = Figure(size = (900, 500))
ga = GeoAxis(
    fig[1, 1]; # any cell of the figure's layout
    dest = "+proj=wintri", # the CRS in which you want to plot
)
lines!(ga, GeoMakie.coastlines()) # plot coastlines from Natural Earth as a reference
scatter!(ga, -120:15:120, -60:7.5:60; color = -60:7.5:60, strokecolor = (:black, 0.2))
b = Colorbar(fig[1,1], tellheight = false, tellwidth = false,
        valign=:bottom, halign=:right,
        width = fig.scene.viewport[].widths[1]/10,
        vertical = false, label = L"\text{GPP} \, (\mu\text{mol m}^{-2} \, \text{s}^{-1})")
save("fig.png", fig)
```

produces a smaller colorbar, with less white space waste, implementation wip

fig

AlexisRenchon:
> Is this relevant only for GeoMakie, or for Makie plots too?

I am not sure what is the default :shading with Makie, but maybe this is relevant for Makie.heatmap2D

AlexisRenchon:
I removed the colorbar from this PR to focus only on no shading, that is definitely needed.

ph-kev (commented):
This looks good to me. I don't think we need it for `heatmap2D!` since I can't find the parameter for no shading in the documentation for `heatmap!` in Makie. 

The only issue is if you call the function like this:
```julia
ClimaAnalysis.Visualize.heatmap2D_on_globe!(
            fig,
            var,
            more_kwargs = Dict(
                :plot => Dict(:title => "My title"),
            ),
        )
```
which will override the default kwargs for that function.

AlexisRenchon:
how can this be avoided?
can it be written in a way that calling the function like your example doesn't override no_shading?

ph-kev:
> how can this be avoided? can it be written in a way that calling the function like your example doesn't override no_shading?

You can do something like
```julia
recursive_merge(x::AbstractDict...) = merge(recursive_merge, x...)
recursive_merge(x...) = x[end]
default_and_more_kwargs = recursive_merge(default_kwargs, more_kwargs)
``` 
and define `default_kwargs` to be
```julia
default_kwargs = Dict(:plot => Dict(:shading => Makie.NoShading))
``` 
and then, use `default_and_more_kwargs` for passing keyword arguments in `Visualize.heatmap2D_on_globe!`. 

The same thing is done in `plot_bias_on_globe!` to get default arguments.

AlexisRenchon:
nice! could you do it? and ask Gabriele for review

Sbozzolo (approved):

AlexisRenchon:
Thanks a lot @ph-kev and @Sbozzolo!! I really appreciate this.

PR #118, opened 2024-10-11
Add masks for land and ocean

### LINKED ISSUE #114: Implement land and ocean mask for computing bias and RMSEs
There is currently no functions that can compute bias and RMSEs over just the land or ocean.

ph-kev:
closes #114  - This PR adds `Var.apply_landmask` and `Var.apply_oceanmask` which sets all the coordinates corresponding to land or ocean to zeros respectively. Every function that compute bias or RMSE now has a mask parameter which take in either mask functions to correctly compute the bias or RMSE.

The plotting functions now use the appropriate mask which solve the problem of masking not affecting the colorbar.

A bias plot with a land mask looks like this:
plot_bias_landmask

A bias plot with an ocean mask looks like this:
plot_bias_oceanmask

Plotting the land mask gives this:
plot_apply_land_mask

Plotting the ocean mask gives this:
plot_apply_ocean_mask


docs/src/var.md
```diff
@@ -255,3 +255,19 @@ julia> global_rmse(sim, obs)
 julia> units(se_var)
 "K^2"
 ```
+
+### Masking
+Bias and squared error can be computed only over the land or ocean though the `mask` parameter.
```
Sbozzolo, line 260:
```suggestion
Bias and squared error can be computed only over the land or ocean through the `mask` parameter.
```

Sbozzolo (approved):
Thank you! Let's also open an issue to update the mask to use whatever we use in the other packages (from ClimaArtifacts) or upload the mask there as well.

PR #141, opened 2024-10-22
Automatically generate a masking function using a `OutputVar`

### LINKED ISSUE #137: Automatically generate masks from OutputVar
For some observational data, data is missing in some places (see plot below). One might be interested in only computing RMSE over regions where the values are defined. Hence, the current land mask would not be enough. It would be helpful to have a function that automatically generate a mask from a `OutputVar` with the longitude and latitude dimensions.

gpp

ph-kev:
closes #137 - This PR adds the function `Var.make_lonlat_mask` which automatically generate a mask from a `OutputVar`. Additionally, one can specify the masking function when plotting the bias. See the plots below where a custom masking function (for computing the bias) is specified with an ocean mask (for visual purposes) for the plot.
plot_bias_with_custom_mask

The `mask` parameter now takes in a masking function or the predefined masks `Visualize.oceanmask()` and `Visualize.landmask()`. If `Visualize.oceanmask()` and `Visualize.landmask()` are used, then the corresponding masking function for the `OutputVar` is chosen. If a masking function is used, then no mask for visual is used. In all other cases, it is assumed that the mask being used is for visuals and a warning is displayed that it will be overplotted and the colorbar might be wrong.


ext/ClimaAnalysisGeoMakieExt.jl
```diff
@@ -380,22 +409,23 @@ function Visualize.plot_bias_on_globe!(
...
     elseif mask == Visualize.landmask()
-        return ClimaAnalysis.apply_landmask
+        return mask, ClimaAnalysis.apply_landmask
     elseif mask == Visualize.oceanmask()
-        return ClimaAnalysis.apply_oceanmask
+        return mask, ClimaAnalysis.apply_oceanmask
     else
         @warn "Mask not recognized, overplotting it. The colorbar will not be correct"
-        return nothing
+        return mask, nothing
```
Sbozzolo, line 428: Can you add a comment to explain the return type?

Sbozzolo (approved):

PR #164, opened 2024-11-14
Add periodic boundary condition when lat is 360°

### LINKED ISSUE #163: Peridoic boundary condition when the longitude dimension is exactly 360 degrees
In our simulation, the longitude dimension ranges from -180 to 180 degrees. As a result, a periodic boundary condition is not made when creating a `OutputVar`.

ph-kev:
closes #163  - When the difference between the longitude dimension is exactly 360.0 degrees, then a periodic boundary condition is added.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -167,6 +167,10 @@ function _find_extp_bound_cond(dim_name, dim_array)
         isapprox(dim_size + dsize, 360.0)
     ) && return Intp.Periodic()
     (
+        conventional_dim_name(dim_name) == "longitude" &&
+        (dim_array[end] - dim_array[begin]) ≈ 360.0
+    ) && return Intp.Periodic()
+    (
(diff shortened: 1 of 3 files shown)


src/Var.jl
```diff
@@ -166,6 +166,10 @@ function _find_extp_bound_cond(dim_name, dim_array)
         _isequispaced(dim_array) &&
         isapprox(dim_size + dsize, 360.0)
     ) && return Intp.Periodic()
+    (
+        conventional_dim_name(dim_name) == "longitude" &&
+        (dim_array[end] - dim_array[begin]) == 360.0
```
Sbozzolo, line 171:
```suggestion
        (dim_array[end] - dim_array[begin]) ≈ 360.0
```
Better account for floating point errors

Sbozzolo (approved):

Issue #165, opened 2024-11-14
Window and slice - using interpolant


ph-kev:
It would be helpful to slice or window a `OutputVar` using arbitrary values rather than the nearest value. It would also be helpful to explicit mention in the documentation that the nearest value is used rather than the value supplied.

PR #176, opened 2024-11-22
Add Var.slice_intp

ph-kev:
This PR adds slicing by using the interpolant of the `OutputVar`.

Helps with issue #165


Sbozzolo (commented):
I think that this should be a keyword argument in `slice`

PR #183, opened 2024-12-04
Add interpolation routine to make `Var.resampled_as` faster

### LINKED ISSUE #182: Var.resampled_as too much allocations
There are a lot of allocations done when resampling. For example, for 3 variables from the ERA5 dataset with 4 dimensions of size (360, 181, 37, 550), the approximate amount being allocated on the heap is nearly 300GB!

ph-kev:
closes #182 - This commit adds an interpolation routine that makes `resampled_as` and any other functions that use interpolations faster.

Checklist
- [x] Add benchmarks on postprocessing pipeline for 4D `OutputVar`s (see [issue](https://github.com/CliMA/ClimaCoupler.jl/pull/1094) on ClimaCoupler)
- [x] Add results from current postprocessing pipeline to demonstrate results do not change
- [x] Add this to NEWS.md

Benchmarks from `@time`
```
# Latest commit on main (on first run)
490.620013 seconds (6.27 G allocations: 370.999 GiB, 5.72% gc time, 9.64% compilation time: 13% of which was recompilation)

# Latest commit on main (on second run)
449.763246 seconds (6.19 G allocations: 365.709 GiB, 6.46% gc time, 0.27% compilation time)

# This PR (on first run)
166.437237 seconds (849.41 M allocations: 151.120 GiB, 7.57% gc time, 28.15% compilation time: 14% of which was recompilation)

# This PR (on second run)
125.102550 seconds (769.52 M allocations: 145.852 GiB, 8.16% gc time, 0.96% compilation time)
``` 

The image below is from the current postprocessing pipeline using the latest commit on main.
bias_pr_all_seasons_orig_intp

The image below is from the current postprocessing pipeline using this PR.
bias_pr_all_seasons_new_intp

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,11 @@ With this release, you can remake a `OutputVar` using an already existing `Outpu
 is helpful if you need to construct a new `OutputVar` from an already existing one, but only
 need to modify one field while leaving the other fields the same.
 
+## Add interpolation routine
+With this release, any functions that rely on interpolation now uses the interpolation
+routine written for ClimaAnalysis instead of Interpolations.jl. This substantially reduce
+the number and size of allocations when using these functions.
+
 v0.5.12
 -------
 
--- a/src/ClimaAnalysis.jl
+++ b/src/ClimaAnalysis.jl
@@ -5,6 +5,7 @@ include("Utils.jl")
 import .Utils
 
 include("Numerics.jl")
+include("Interpolations.jl")
 
 include("Var.jl")
 @reexport using .Var
--- a/src/Interpolations.jl
+++ b/src/Interpolations.jl
@@ -1,0 +1,276 @@
+module Interpolations
+
+#=
+Why not use Interpolations.jl?
+
+One of the most expensive operation in terms of time and memory is
+`Var.resampled_as(src_var, dest_var)` which uses an interpolation to resample `data` in
+`src_var` to match the `dims` in `dest_var`. However, interpolating using Interpolations.jl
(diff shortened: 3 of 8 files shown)


src/Numerics.jl
```diff
@@ -166,4 +166,248 @@ function _integration_weights_generic_equispaced(arr)
...
+
+function linear_interpolate(
+    point::Tuple,
+    axes,
+    data::AbstractArray{FT2, N},
+    extp_conds,
+) where {N, FT2}
+    point = promote(point...)
+    return linear_interpolate(point, axes, data, extp_conds)
+end
```
Sbozzolo, line 258: Can you add comments to explain what these methods are accomplishing?

src/Numerics.jl
```diff
@@ -166,4 +166,248 @@ function _integration_weights_generic_equispaced(arr)
...
+    end, N)
+end
+
+"""
+    get_indices(indices::NTuple{N, Tuple{I, I}}, bits) where {N, I}
+
+Given a tuple consisting of 2-tuple, return a tuple of one element from each tuple according to bits.
+"""
+function get_indices(indices::NTuple{N}, bits) where {N}
+    return ntuple(dim -> if (bits & (1 << (dim - 1)) != 0)
```
Sbozzolo, line 303: Can you add a link to where you found this bit manipulation?

Sbozzolo (commented):
Could you also move all the linear interpolation into its own module, so that it will be easier in the future to switch it to something else?

ph-kev:
This PR is not needed with https://github.com/CliMA/ClimaAnalysis.jl/pull/418 merged.

PR #188, opened 2025-01-24
Add check for whether P(z) is bijective or not

ph-kev:
This PR adds a check for whether P(z) is bijective or not when resampling columns.

### DIFF
--- a/src/Atmos.jl
+++ b/src/Atmos.jl
@@ -22,11 +22,17 @@ Linearly interpolate `var1d` from `origin_pressure` to `target_pressure`.
 Note: Values outside of the range are linearly extrapolated
 """
 function _resample_column!(dest, var1d, origin_pressure, target_pressure)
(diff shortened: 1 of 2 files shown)


src/Atmos.jl
```diff
@@ -22,6 +22,8 @@ Linearly interpolate `var1d` from `origin_pressure` to `target_pressure`.
 Note: Values outside of the range are linearly extrapolated
 """
 function _resample_column!(dest, var1d, origin_pressure, target_pressure)
+    !issorted(origin_pressure, lt = !isless) &&
```
Sbozzolo, line 25: There's many negative in this line and it is a little opaque. Do you need `!isless` to account for identical values?

ph-kev:
There is a new function `allunique` that was released in Julia 1.11, but I think it would fail the downgrade tests if I use it here.

Sbozzolo (approved):

PR #195, opened 2025-02-13
fix longitude_name call in apply_latlon_mask

### LINKED ISSUE #191: `make_latlon_mask` fails if longitude not named "lon"
`make_latlon_mask` indexes into `input_var` using the longitude/latitude names of `mask_var`, which are set in that function to be "lon" and "lat". This fails if `input_var` uses different latitude/longitude names (e.g. "latitude", "longitude"). This function should instead use the lon/lat names of `input_var` when indexing into it.

https://github.com/CliMA/ClimaAnalysis.jl/blob/28823aa260bb2768c95289b8ef2b77fe3ee5c69c/src/Var.jl#L1785-L1786

juliasloan25:
## Purpose 
closes #191 

The test I added fails without the change in src, and passes with it

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -1782,8 +1782,8 @@ function make_lonlat_mask(
         # Reshape data for broadcasting
         lon_idx = input_var.dim2index[longitude_name(input_var)]
         lat_idx = input_var.dim2index[latitude_name(input_var)]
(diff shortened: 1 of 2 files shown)


Sbozzolo (approved):
Thank you!

PR #203, opened 2025-02-27
Fix Var.weighted_average_lat bug with NaN and FT32

ph-kev:
In the function `Var.weighted_average_lat`, if latitudes is in `Float32` and there are `NaN`s in the data, then eventually the computation being done is `Vector{Real}` times `Vector{Float32}` which gives `Vector{Real}`. Then, `nanmean` is computed on this vector, but throw an error since there is no function defined for that type.


AlexisRenchon (approved):
Thank you Kevin!

Sbozzolo (approved):
Feel free to merge after you address the comment above

PR #206, opened 2025-02-28
Add valid_time to TIME_NAMES and fix issue with pretty printing

### LINKED ISSUE #204: Pretty print when dictionaries are empty throw an error
When the dictionaries of the `OutputVar` are empty, then printing will throw an error when trying to compute the amount of padding to add.

There should be tests that test for the empty OutputVar and when the dictionaries are empty.

### LINKED ISSUE #205: Add compatibility for reading NetCDF files from ERA5
The time dimension is called `valid_time` in the NetCDF file. The name should also be in `ClimaAnalysis.Var.TIME_NAMES`.

ph-kev:
closes #204, closes #205 - The issue with pretty printing is when the `OutputVar` has empty dictionaries in which the padding cannot be computed. The other potential issue is when the value of `dim_attributes` is something that is not a dictionary.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -1858,7 +1858,8 @@ function Base.show(io::IO, var::OutputVar)
     # Print the key value pairs of attributes
     printstyled(io, "Attributes:\n", bold = true, color = :green)
     # Find spacing to pad out key of var.attributes
-    max_length_attribs = maximum(length(x) for (x, _) in var.attributes)
+    max_length_attribs =
+        maximum(length(x) for (x, _) in var.attributes; init = 0)
     for (key, val) in var.attributes
         print(io, "  " * rpad(key, max_length_attribs) * " => ")
         printstyled(io, "$(val)\n", color = :light_cyan)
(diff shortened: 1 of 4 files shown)


Sbozzolo:
I don't think there should be a case where `dim_attributes` is not a dict. It's type is `OrderedDict{String, C}` in the type definition

ph-kev:
> I don't think there should be a case where `dim_attributes` is not a dict. It's type is `OrderedDict{String, C}` in the type definition

I realized that I poorly wrote that sentence. I mean that `C` can be something that is not a dictionary.

ph-kev:
> > > I don't think there should be a case where `dim_attributes` is not a dict. It's type is `OrderedDict{String, C}` in the type definition
> > 
> > 
> > I realized that I poorly wrote that sentence. I mean that `C` can be something that is not a dictionary.
> 
> Are we encountering these cases in the wild? Which other `C`s do we have to deal with?

I don't think this ever appear in the wild, but we don't restrict what `C` could be in the struct for an `OutputVar`. I agree that there shouldn't be a case where `dim_attributes` is not a `Dict`. Should I add the constraint for what `C` is in the `OutputVar`?

src/Var.jl
```diff
@@ -1897,14 +1901,19 @@ function Base.show(io::IO, var::OutputVar)
...
             end
-        end
-        if length(array) == 1
+        elseif length(array) == 1
             print(io, "element")
             print(io, " (")
             printstyled(io, "$(array[begin])", color = :light_cyan)
             print(io, ")")
+        else
+            length(array) == 0
```
Sbozzolo, line 1910:
```suggestion
        elseif length(array) == 0
```

Sbozzolo (approved):

PR #207, opened 2025-02-28
Add Var.split_by_season_across_time

ph-kev:
This PR adds `Var.split_by_season_across_time`. The difference between this and `Var.split_by_season` is that the splitting is also done across time. 

- [ ] Add to NEWS.md
- [ ] Add documentation


NEWS.md
```diff
@@ -1,5 +1,12 @@
 ClimaAnalysis.jl Release Notes
 ===============================
+v0.5.14
+-------
+
+## Split by seasons across time
+It may be the case that you want to split seasons, but also want to retain the order that
+the seasons appear in across time. This can be done by using `split_by_season_across_time`.
```
Sbozzolo, line 8: I don't understand this. Could you add an example?

src/Utils.jl
```diff
@@ -304,6 +305,89 @@ function split_by_season(dates::AbstractArray{<:Dates.DateTime})
     return (MAM, JJA, SON, DJF)
 end
 
+"""
+    split_by_season_across_time(dates::AbstractArray{<:Dates.DateTime})
+
+Split `dates` into a vector of vectors of dates split by seasons across time.
```
Sbozzolo, line 311: Can you add some further description here?

It is not clear what this function does just by reading the docstring

src/Utils.jl
```diff
@@ -304,6 +305,89 @@ function split_by_season(dates::AbstractArray{<:Dates.DateTime})
...
+ [DateTime("2010-12-01T00:00:00")]
+"""
+function split_by_season_across_time(dates::AbstractArray{<:Dates.DateTime})
+    # Dates is not necessarily sorted
+    dates = sort(dates)
+
+    # Empty case
+    if length(dates) == 0
+        return Vector{Vector{eltype(dates)}}[]
+    end
```
Sbozzolo, line 336:
```suggestion
    isempty(dates) && return Vector{Vector{eltype(dates)}}[]
```
(Or in general, use `isempty` instead of checking the length so that it is a little bit more clear)

src/Var.jl
```diff
@@ -1297,41 +1299,97 @@ expected to be second. Also, the interpolations will be inaccurate in time inter
...
+    season_times =
+        (date_to_time.(start_date, season) for season in season_dates)
+
+    return _split_along_dim(var, time_name(var), season_times)
+end
+
+"""
+    split_by_season_across_time(var::OutputVar)
+
+Return a vector of `OutputVar`s split by season across time.
```
Sbozzolo, line 1315: Can you add some further description here?

It is not clear what this function does just by reading the docstring and how it differs from `split_by_season`

src/Var.jl
```diff
@@ -1297,41 +1299,97 @@ expected to be second. Also, the interpolations will be inaccurate in time inter
...
+
+The months of the seasons are March to May, June to August, September to November, and
+December to February. If there are no dates found for a season, then the `OutputVar` for
+that season will be an empty `OutputVar`. The first `OutputVar` is guaranteed to not be
+empty.
```
Sbozzolo, line 1320:
```suggestion
The months of the seasons are March to May (MAM), June to August (JJA), September to November (SON), and
December to February (DJF). If there are no dates found for a season, then the `OutputVar` for
that season will be an empty `OutputVar`. The first `OutputVar` is guaranteed to not be
empty.
```

src/Var.jl
```diff
@@ -1297,41 +1299,97 @@ expected to be second. Also, the interpolations will be inaccurate in time inter
...
+
+The function will use the start date in `var.attributes["start_date"]`. The unit of time is
+expected to be second. Also, the interpolations will be inaccurate in time intervals
+outside of their respective season for the returned `OutputVar`s.
```
Sbozzolo, line 1324: Let's move this "Also" in a note and explain it a little bit better (maybe with an example)

Sbozzolo (commented):
Can you please also add a FAQ explaining the difference between the two functions and caveats?

Sbozzolo (approved):
Thank you!

PR #216, opened 2025-03-10
Fix Var.center_longitude!

ph-kev:
This commit adds `Var.shift_longitude` which replace `Var.center_longitude!`. 

## TODO
- [x] Add tests
- [x] Add to NEWS.md
- [x] Add an example of how to use it in docs if one doesn't already exist

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -35,6 +35,7 @@ export OutputVar,
     slice,
     window,
     arecompatible,
+    shift_longitude,
     center_longitude!,
     short_name,
     long_name,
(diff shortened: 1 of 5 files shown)


src/Var.jl
```diff
@@ -1053,8 +1054,16 @@ end
...
+
+!!! warn "Deprecated"
+    This function is deprecated and users are encouraged to use
+    [`shift_longitude`](@ref) instead.
 """
 function center_longitude!(var, lon)
+    @warn(
+        "This function is deprecated and may be incorrect. Users are encouraged to use
+  `center_longitude(var, lower_lon, upper_lon; shift_by = 0.0)` instead."
+    )
```
Sbozzolo, line 1066: There's a built-in way to do deprecation in Julia:

https://invenia.github.io/blog/2022/06/17/deprecating-in-julia/

src/Var.jl
```diff
@@ -1078,6 +1087,62 @@ function center_longitude!(var, lon)
...
+    if (lon[end] - lon[begin] ≈ 360.0)
+        pop!(lon)
+    end
+
+    # Center longitudes
+    for i in eachindex(lon)
+        if (lon[i] >= upper_lon) || (lon[i] < lower_lon)
+            lon[i] = (lon[i] - lower_lon + width) % width + lower_lon
+        end
+    end
```
Sbozzolo, line 1131: I wonder if this is more clear:

```suggestion
function wrap_longitude(long, lower, upper, width)
    return (long >= upper || long < lower) ? ((long - lower + width) % width + lower) : long
end

lon .= wrap_longitude.(lon, lower_lon, upper_lon, width)
```

Sbozzolo: In general, could you add two lines of comment to describe the high level approach taken in this function?

Sbozzolo (approved):

PR #247, opened 2025-04-10
Add support for dates in `slice` and `window`

Sbozzolo:
`slice` and `window` are very commonly used functions. One of the main use cases it selecting a moment of time to plot, or a subset of data to analyze.

When working with real observations, it is often very convenient to operate at the level of dates, instead of simulation time. This PR adds support for this feature.

Now, the following operations are supported
```julia
slice(var, time = Dates.DateTime(2009, 12, 15))
window(var, left = Dates.DateTime(2009, 12, 15), right = Dates.DateTime(2009, 12, 17))
```

The requirement for this is that `var` contains a `start_date` among its attributes. We also assume that the units of `time` are second.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,23 @@ Now, `slice` and `window` will try to match the name of the dimension with a
 database of known names. For example, they understand `lat` even if the data has
 dimensions `latitude`.
 
+`slice` and `window` can also now work directly with `Dates.DateTime`s (assuming
+that the `OutputVar` contains enough information to convert from `times` to
+`dates`, namely the `start_date` among the attributes).
+
+Example
+```julia
+# Suppose `var` is an `OutputVar`
+julia> var.attributes["start_date"]
+2008-01-01T00:00:00
+
+julia> time_name(var)
+"t"
+
+# Notice how we call `time` and not `t`. This was not previously possible!
+julia> slice(var, time = Dates.DateTime(2009, 12, 15))
+```
+
 v0.5.16
 -------
 
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -639,7 +639,7 @@ end
 """
     remake(var; attributes, dims, dim_attributes, data)
 
-remake a `OutputVar`. If a keyword argument is not supplied, then it defaults
+Remake an `OutputVar`. If a keyword argument is not supplied, then it defaults
 to `attributes`, `dims`, `dim_attributes`, or `data` of `var`.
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -11,10 +11,27 @@ This is fixed in this release.
 
 ## More flexible `slice` and `window`
 
-Now, `slice` and `window` will try to match the name of the dimension with a
+nNow, `slice` and `window` will try to match the name of the dimension with a
```
ph-kev, line 14:
```suggestion
Now, `slice` and `window` will try to match the name of the dimension with a
```

src/Var.jl
```diff
@@ -1262,15 +1312,31 @@ Return a new OutputVar by selecting the values of the given `dim`ension that are
...
 """
 function window(var, dim_name; left = nothing, right = nothing)
     dim_name_in_var = find_corresponding_dim_name_in_var(dim_name, var)
 
+    dim_is_time = conventional_dim_name(dim_name_in_var) == "time"
+
+    # If val is a Date, let's compute the associated time
+    _maybe_convert_to_time(num) =
+        num isa Dates.DateTime ? date_to_time(var, num) : num
+    left, right = _maybe_convert_to_time(left), _maybe_convert_to_time(right)
```
ph-kev, line 1338: The variable `dim_is_time` is never used. I imagine that it should be used since you can get a case where someone put a date, but accidentally use a dimension that isn't time.

NEWS.md
```diff
@@ -11,10 +11,27 @@ This is fixed in this release.
...
+# Suppose `var` is an `OutputVar`
+julia> var.attributes["start_date"]
+2008-01-01T00:00:00
+
+julia> time_name(var)
+"t"
+
+# Notice how we call `time` and not `t`. This was not previously possible!
+julia> slice(var, time = Dates.DateTime(2009, 12, 15))
+```
```
ph-kev, line 33: Could you use `@repl` instead? I don't want the example to become stale and break if this ever get updated.

ph-kev (changes requested):

NEWS.md
Sbozzolo: This is inside the NEWS markdown file, so `julia` here is just for syntax highlighting

src/Var.jl
Sbozzolo: Added checks

ph-kev (approved):

PR #253, opened 2025-04-11
Try returning dates when possible

Sbozzolo:
When we have a `start_date`, we can try transforming from `times` to `dates`. This is useful when one wants to inspect and manipulate Vars at the level of calendar dates


NEWS.md
```diff
@@ -9,6 +9,13 @@ v0.5.17
 resampling may produce incorrect results when interpolating outside of the longitude range.
 This is fixed in this release.
 
+## Reconstructing dates
+
+`ClimaAnalysis.dates` will now try returning a list of `Dates` even when `dates` is not
+among the available dimensions. For this to work, `var.attribtues` have to contain
```
ph-kev, line 15:
```suggestion
among the available dimensions. For this to work, `var.attributes` have to contain
```

src/outvar_dimensions.jl
```diff
@@ -136,8 +136,25 @@ date_name(var::OutputVar) = find_dim_name(keys(var.dims), DATE_NAMES)
...
+`times` is seconds.
 """
-dates(var::OutputVar) = var.dims[date_name(var)]
+function dates(var::OutputVar)
+    has_date(var) && return var.dims[date_name(var)]
+
+    # Assuming time is in seconds
+    if has_time(var) && haskey(var.attributes, "start_date")
+        return Dates.DateTime(var.attributes["start_date"]) .+
+               Dates.Second.(times(var))
```
ph-kev, line 151: Can we use `date_to_time` in `Utils` instead?

ph-kev (changes requested):

NEWS.md
Sbozzolo: Fixed

src/outvar_dimensions.jl
Sbozzolo: Yes

ph-kev (approved):

PR #256, opened 2025-04-14
Add identity

Sbozzolo:
This is useful when we need a no-op function, but still want to apply a function (as when preprocessing some OutputVars).


src/Var.jl
```diff
@@ -2534,6 +2534,10 @@ function Base.show(io::IO, var::OutputVar)
     end
 end
 
+function Base.identity(var::OutputVar)
+    return var
+end
```
ph-kev, line 2539: Should there be documentation for this function?

ph-kev (approved):

src/Var.jl
Sbozzolo: I can add it

PR #267, opened 2025-05-02
Add Base.cat

### LINKED ISSUE #264: Combine `OutputVar`s that share dimensions except for time
It would be convenient to be able to combine multiple `OutputVar`s that share dimensions except for time. The only current way to do this is with `remake`, ensuring that the input data and dimensions have been set properly. 

This feature would enable me to extract a specific season over many years from an `OutputVar` that spans several years:
```julia
ql_seasons = split_by_season_across_time(ql)
ql_just_springs = combine(ql_seasons[1:4:end])
```

This could also be extended to support concatenation for `OutputVar`s that share all dimensions but one.

ph-kev:
closes #264 - This PR adds `Base.cat` for `OutputVar`s and also make `dim_units` agnostic to the name of the dimension.

Only the start date, sort name, and units are preserved are kept in the resulting `OutputVar`.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -2638,6 +2638,120 @@ function reverse_dim!(var::OutputVar, dim_name)
 end
 
 """
+    Base.cat(vars::OutputVar...; dims::String)
+
+Concatenate `vars` along the dimension `dims`.
+
+This function is helpful if you need to concatenate `OutputVar`s after applying
+`split_by_season_across_time` or `window`.
+
+This function does not support concatenating `OutputVar`s of different quantities as
+determined by the `short_name`.
+
+Attributes that are not `start_date`, `short_name`, or "units" are discarded in the process.
+
+Note that concatenation along multiple dimensions is not possible.
+"""
+function Base.cat(vars::OutputVar...; dims::String)
+    last_var = last(vars)
+    vars_dims = (var.dims for var in vars)
+
+    # Find the index of `dim`
+    dim_idx =
+        last_var.dim2index[find_corresponding_dim_name_in_var(dims, last_var)]
+
+    # Check the number of dimensions for all vars
+    all(var -> length(var.dims) == length(last_var.dims), vars) ||
+        error("The number of dimensions are not the same across all OutputVars")
+
+    # Check order of dimensions are the same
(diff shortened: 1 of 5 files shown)


src/Var.jl
```diff
@@ -2638,6 +2637,115 @@ function reverse_dim!(var::OutputVar, dim_name)
...
+`split_by_season_across_time` or `window`.
+
+This function does not support concatenating `OutputVar`s of different quantities as
+determined by the `short_name`.
+
+Attributes that are not `start_date`, `short_name`, or "units" are discarded in the process.
+
+Note that concatenation along multiple dimensions is not possible.
+"""
+function Base.cat(vars::OutputVar...; dim::String)
```
ph-kev, line 2655: I wasn't too sure about whether is should be `dim` or `dims`, since Base.cat has `dims`. On the other hand, this function only support concatenating along a single dimension.

ph-kev: If it is `dim` and the user forgot to make it a keyword argument, then the error message is not helpful since it defaults to `Base.cat`.

src/Var.jl
```diff
@@ -2638,6 +2637,115 @@ function reverse_dim!(var::OutputVar, dim_name)
...
+    for attrib in keep_attribs
+        all(
+            var ->
+                get(var.attributes, attrib, nothing) ==
+                get(last_var.attributes, attrib, nothing),
+            vars,
+        ) || error("Not all OutputVars have the same $attrib")
+    end
+
+    # Keep attributes of the last OutputVar There are multiple approaches for choosing which
```
Sbozzolo, line 2710:
```suggestion
    # Keep attributes of the last OutputVar. There are multiple approaches for choosing which
```

src/Var.jl
```diff
@@ -2638,6 +2637,115 @@ function reverse_dim!(var::OutputVar, dim_name)
...
+    # attributes to keep, such as merging the attributes recursively, taking an intersection
+    # of the attributes, or keeping the attributes of the last OutputVar We take a
```
Sbozzolo, line 2712:
```suggestion
    # of the attributes, or keeping the attributes of the last OutputVar. We take a
```

Sbozzolo (approved):

PR #269, opened 2025-05-07
Add FlatVar

### LINKED ISSUE #240: Add function to flatten OutputVar
It might be useful to implement something like `flatten(var; dims)`, where you can specify the order of the dimensions to flatten the `OutputVar.` I am unsure if the result should be another `OutputVar`, but it would be nice if it is. If the result is an `OutputVar`, then implementing the function `unflatten` would also be helpful.

A function to look at that might be useful for implementing is `Var.reordered_as`.

ph-kev:
closes #240 - This PR adds functionality to help with calibration by giving a consistent way of flattening an `OutputVar` without worrying about using `Base.permutedims` everywhere.

This is still a work in progress, because this has not been tested with any actual calibration attempt. Furthermore, there is still the question of how to handle operations where masking is needed.

For example, with ClimaLand calibration, you need to only consider points over land. Let say that you want to visualize the observations from the EKP object. As of now, there is not enough information to reconstruct the `OutputVar` without also keeping track of which values are `NaN`.

### DIFF
--- a/src/flat.jl
+++ b/src/flat.jl
@@ -1,0 +1,195 @@
+export flatten, unflatten, flatten_dim_order
+
+"""
+    Representing the metadata of an `OutputVar` and contain all the necessary
+    information to reconstruct an `OutputVar`.
+"""
+struct Metadata{T <: AbstractArray, B, C, I <: Integer}
+    "Attributes associated to this variable, such as short/long name"
+    attributes::Dict{String, B}
+
+    "Dimensions over which the variable is defined"
+    dims::OrderedDict{String, T}
+
+    "Attributes associated to the dimensions"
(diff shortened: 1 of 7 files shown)


src/flat.jl
```diff
@@ -0,0 +1,188 @@
...
+
+    "Indices of NaNs in data"
+    nan_indices::Vector{I}
+end
+
+"""
+    Representing a flat output variable
+"""
+struct FlatVar{A <: AbstractVector}
+
```
Sbozzolo, line 28:
```suggestion
```

Sbozzolo (approved):
Look good! Thank you!

Could you also add a NEWS entry?

If the news file is getting too big, we can move part of it to a HISTORY file

PR #276, opened 2025-05-19
Add Var.average_seasons_across_time

ph-kev:
This PR adds `Var.average_seasons_across_time`, which computes the seasonal average of an `OutputVar` and returns a single `OutputVar`. In addition, the attributes `year` and `season` are stored in the returned `OutputVar` as vectors of years and seasons, respectively.

### DIFF
--- a/src/Utils.jl
+++ b/src/Utils.jl
@@ -346,6 +346,9 @@ function split_by_season_across_time(dates::AbstractArray{<:Dates.DateTime})
     # Find the first date of the season that first(dates) belongs in
     (first_season, first_year) = find_season_and_year(first(dates))
     season_to_month = Dict("MAM" => 3, "JJA" => 6, "SON" => 9, "DJF" => 12)
+    # Because the year is determined from the second month, we need to handle the case when
+    # the season is DJF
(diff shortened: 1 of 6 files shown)


src/Utils.jl
```diff
@@ -431,21 +431,32 @@ Return a tuple of the year and season belong to `date`. The variable `year` is
...
 
 The months of the seasons are March to May, June to August, September to
-November, and December to February. If a date is in December to February, the
-year is chosen to be the year that the season starts.
+November, and December to February.
+
+If `ref_month` is `:first` and a date is in December to February, then the first
+month of each season is used to compute what year the season is. If `ref_month`
+is `:second` and a date is in December to February, then the second month of
+each season is used to compute what year the season is.
```
Sbozzolo, line 439: Do you need this? 

I would think that we alwasys want DJF to be associated to the year of JF

ph-kev: It is used in `Utils.split_by_season_across_time`, but I can change it there instead of in `find_season_and_year`.

Sbozzolo (approved):

PR #282, opened 2025-05-27
Exclude NaNStatistics v0.6.51 and v0.6.52 in compat

ph-kev:
This commit upper bounds NaNStatistics to v0.6.50. This is because in v0.6.51 and v0.6.52, there is a bug with NaNStatistics, where `nanmean` is not correct. See the issue:
https://github.com/brenhinkeller/NaNStatistics.jl/issues/59.


ph-kev:
I am going to also add some text to NEWS.md about why those versions are excluded
.

PR #285, opened 2025-06-05
Add add_data and add_data! to developer docs

ph-kev:
This commit also made small changes to the documentation for the other functions that add data to `TemplateVar`.

### DIFF
--- a/docs/src/developer.md
+++ b/docs/src/developer.md
@@ -78,6 +78,8 @@ Template.add_attribs
 Template.add_attribs!
 Template.add_dim
 Template.add_dim!
+Template.add_data
+Template.add_data!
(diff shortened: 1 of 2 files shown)


ph-kev:
> I think it would be nice to be clearer about what "designed to be used with the pipe operator" means. Without reading the content of the functions, I would expect `add_data` to create a TemplateVar and the pass that to `add_data!`, not return an anonymous function. Same for `one_to_n_data`

I was more of intending for whoever is going to use this to read the documentation instead of the API first, since with an example, it is a lot more clearer what I mean by using the pipe operator.

imreddyTeja (approved):

PR #295, opened 2025-06-30
Support reading multiple files into a OutputVar

### LINKED ISSUE #294: Add functionality to load an OutputVar from multiple files
This feature is already implemented for loading `OutputVar` from `SimDir`.

ph-kev:
closes #294 - This PR supports reading multiple files into a `OutputVar`.


make_nc.jl
imreddyTeja, line 1: Did you intentionally put this file in the main directory?

imreddyTeja (changes requested):
Looks good, but I think you accidentally added a file. This should also be announced in NEWS.nd

src/Var.jl
```diff
@@ -258,12 +258,12 @@ function OutputVar(dims, data)
...
 
 """
-    OutputVar(path,
+    OutputVar(path::Union{String, Vector{String}},
               short_name = nothing;
               new_start_date = nothing,
               shift_by = identity)
 
-Read the NetCDF file in `path` as an `OutputVar`.
+Read the NetCDF file(s) in `path` as a `OutputVar`.
```
imreddyTeja, line 266: I think it would be good to specifically state that if path is a vector, it should be files with the same var and dims (other than time), and they should be sorted by time.

imreddyTeja (approved):
Looks good. I left a comment about making the behavior of passing in a vector of paths explicitly clear in the docstring

imreddyTeja (approved):

PR #297, opened 2025-07-08
Make threshold a keyword argument

ph-kev:
The argument `threshold` should be a keyword argument and not a positional argument.

### DIFF
--- a/src/masks.jl
+++ b/src/masks.jl
@@ -110,7 +110,7 @@ function generate_lonlat_mask(
 end
 
 """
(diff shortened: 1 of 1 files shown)


ph-kev:
> What's the reasoning behind making is a kwarg vs a positional arg? This is a question I have in general because there is nothing in the style guide on when to use kwargs vs positional args.

For ClimaAnalysis, I don't think we have anything specific for when to use keyword argument or positional arguments. For this specific case, it should be a keyword argument because the other masking functions use `threshold` as a keyword argument.

PR #302, opened 2025-07-24
Add NCCatalog

ph-kev:
This PR adds `NCCatalog` for organizing and loading `OutputVar` from NetCDF files. This is the first step towards cleaning up the how data is loaded from NetCDF files and for processing them as the interface between `SimDir` and `NCCatalog` should mostly be the same.

TODO
- [x] Add to NEWS.md

### DIFF
--- a/src/Catalog.jl
+++ b/src/Catalog.jl
@@ -1,0 +1,140 @@
+module Catalog
+
+import NCDatasets
+
+import Base: get
+import ..Sim: available_vars
+import ..Var: OutputVar
+
+export NCCatalog, add_file!, available_vars
+
+"""
+    Organize external NetCDF files to easily initialize `OutputVar`s.
+"""
+struct NCCatalog
+    "Dictionary that maps aliases to tuples of file path and variable name in the NetCDF
+    file. The identifier can either be the variable name or an alias for the variable name
+    in the file."
+    identifier_to_path_varname::Dict{String, Tuple{String, String}}
+
+    "Dictionary that maps variable names to filepaths. This is a fallback if a variable is
+    requested and cannot be found in `identifier_to_path_varname`."
+    varname_to_filepath::Dict{String, String}
+end
+
+"""
+    NCCatalog()
+
+Construct an empty `NCCatalog`.
+
+Files can be added with [`ClimaAnalysis.Catalog.add_file!`](@ref).
+"""
+function NCCatalog()
+    identifier_to_path_varname = Dict{String, Tuple{String, String}}()
+    varname_to_filepath = Dict{String, String}()
+    return NCCatalog(identifier_to_path_varname, varname_to_filepath)
+end
+
+"""
+    add_file!(catalog::NCCatalog, filepath, short_names...)
+
+Add the file at `filepath` to `catalog` with `short_names`.
+
+The argument `short_names` can either be a short name (e.g., `"rsdt"`) or a pair of strings
+(e.g., `"solar_mon" => "rsdt"`). If a pair is used, then the first name is the variable name
+in the NetCDF file and the second name is the alias which will be used for `get`.
+
+!!! note "Multiple files with the same short names"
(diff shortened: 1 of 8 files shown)


src/Catalog.jl
```diff
@@ -0,0 +1,137 @@
...
+(e.g., `"solar_mon" => "rsdt"`). If a pair is used, then the first name is the variable name
+in the NetCDF file and the second name is the alias which will be used for `get`.
+
+!!! note "Multiple files with the same short names"
+    If there are multiple files that contain the same short names, then you should pass in
+    the desired file first or pass in a short name or a pair of short name to alias for
+    `short_names` to make the variable in the file retrievable with `get`.
+"""
+function add_file!(catalog::NCCatalog, filepath, short_names...)
+    isfile(filepath) || error("$filepath does not exist")
```
nefrathenrici, line 53: Nitpick for a slightly clearer error:
```suggestion
    throw(ArgumentError("NetCDF file not found: $filepath"))
```

src/Catalog.jl
```diff
@@ -0,0 +1,137 @@
...
+import Base: get
+import ..Sim: available_vars
+import ..Var: OutputVar
+
+export NCCatalog, add_file!, available_vars
+
+"""
+    Organize external NetCDF files to easily initialize `OutputVar`s.
+"""
+struct NCCatalog{DOI <: Dict, DOV <: Dict}
```
nefrathenrici, line 14: Nitpick, you could specify the parameterized types for `DOI` and `DOV` here rather it in the constructor: `Dict{String, String}` and `Dict{String, Tuple{String, String}}`

src/Catalog.jl
```diff
@@ -0,0 +1,137 @@
...
+Return a `OutputVar` with `short_name` from `catalog`.
+
+The keyword argument `var_kwargs` is a tuple of keyword arguments passed to `OutputVar`. See
+[`ClimaAnalysis.Var.OutputVar`](@ref) for a list of keyword arguments that can be passed.
+
+!!! note "Multiple files with the same short name"
+    If there are multiple files that contain the same `short_name`, then the returned
+    `OutputVar` will be from the NetCDF file with the short name or alias specified, or from
+    the file that was added first if no short names were provided when adding the files.
+"""
```
nefrathenrici, line 98: Could you specify the fallback behavior in the docstring as well?

nefrathenrici (approved):
Thank you! I just had a few nitpick comments. It would be nice to add filepath globbing support in a future PR:
`ClimaAnalysis.add_file!(catalog, "temperature_*.nc")`

src/Catalog.jl
```diff
@@ -0,0 +1,137 @@
...
+end
+
+"""
+    available_vars(catalog::NCCatalog)
+
+Return the short names of the variables found in `catalog`.
+
+Both aliases and variable names in the NetCDF files are returned.
+"""
+function available_vars(catalog::NCCatalog)
```
imreddyTeja, line 122: This is probably not worth it because `simdir` doesn't have a method for it, but I think it might make sense to have a `keys(::NCCatalog)` method because it matches the behavior of similar Julia types

imreddyTeja (commented):
Does `in(::String, ::NCCatalog)` work?

src/Catalog.jl
ph-kev: I am confused by this, because `available_vars` is implemented for `SimDir`?

imreddyTeja: Yeah sorry that comment is unclear. I was trying to say that this function seems like it should be names `keys`, because that is how you would get the varnames from an NCDataset or dictionary. But it is probably best to follow the pattern implemented by `SimDir`

ph-kev:
> Does `in(::String, ::NCCatalog)` work?

No. I don't think it should be implemented since it is not implemented for `SimDir`, and I think the interfaces for both should remain mostly the same.

imreddyTeja (approved):

PR #309, opened 2025-08-04
Add accessor functions for Metadata and FlatVar

### LINKED ISSUE #296: Add accessor functions for metadata
In ClimaCalibrate, it should be possible to use the metadata to check for the units, dates, and grid. To accomplish, accessor functions (e.g. `short_name`, `longitudes`, `dates`) should be made to work with metadata as well.

One way to accomplish this is to take the metadata, make it into a `OutputVar` with any data, and pass it to the correct function. This assumes that making a `OutputVar` is not too much overhead.

ph-kev:
closes #296 - This PR overloads `propertynames` and `get_property`, so that `dims`, `dim_attributes`, and `attributes` can be accessed. Furthermore, accessor functions like `dates` and `has_time` can be used for `FlatVar` and `Metadata`.

We use a Union type instead of the Holy Traits design pattern for simplicity, as traits would require additional dispatch methods for each function without significant benefit for this use case. For example, the user would most likely not use the trait themselves.

# TODO

- [x] Add to NEWS.md
- [x] Add documentation
- [x] Write tests for accessors
- [x] Add tests for propertynames
- [x] Come up with a better name for HasDim

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -98,6 +98,15 @@ struct OutputVar{T <: AbstractArray, A <: AbstractArray, B, C <: AbstractDict}
     index2dim::Vector{String}
 end
 
+# Define FlatVar and Metadata
+include("flat.jl")
+
+# Union type for objects that contains the fields `attributes`, `dims`, and
+# `dim_attributes`. The implementation of these fields must be identical to
+# the fields for `OutputVar`. This union type is needed to make functions that
+# only use these fields work for `OutputVar`, `FlatVar`, and `Metadata`.
+const HasDimAndAttribs = Union{OutputVar, FlatVar, Metadata}
+
 """
     _make_interpolant(dims, data)
 
@@ -467,7 +476,7 @@ Return the `short_name` of the given `var`, if available.
 
 If not available, return an empty string.
 """
-function short_name(var::OutputVar)
+function short_name(var::HasDimAndAttribs)
     get(var.attributes, "short_name", "")
 end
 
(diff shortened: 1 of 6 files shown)


src/flat.jl
```diff
@@ -193,3 +193,32 @@ function _data_size(metadata::Metadata; ignore_nan = true)
...
+
+"""
+    Base.propertynames(var::FlatVar)
+
+Return the property names for `FlatVar` which includes both the actual fields (`metadata`,
+`data`) and the virtual properties (`attributes`, `dims`, `dim_attributes`).
+"""
+function Base.propertynames(var::FlatVar)
+    return (:metadata, :data, :attributes, :dims, :dim_attributes)
+end
```
ph-kev, line 224: You could say that the details of attributes, dims, and dim_attributes should be kept internal, since it is confusing to understand and an user should not really need to use it. For example, is the order of `dims` before or after flattening?

src/Var.jl
```diff
@@ -98,6 +98,13 @@ struct OutputVar{T <: AbstractArray, A <: AbstractArray, B, C <: AbstractDict}
     index2dim::Vector{String}
 end
 
+# Define FlatVar and Metadata
+include("flat.jl")
+
+# Union type for objects that have dimension information but don't require data
+# access
+const HasDim = Union{OutputVar, FlatVar, Metadata}
+
```
ph-kev, line 107: This isn't a good name, since we only care about the fact that we do not need to access the data for these functions.

ph-kev: > I'm not sure what you mean by " we do not need to access the data". Does this mean that certain properties of the data is accessible without accessing the data directly?

It means that we don't need to access the data field of the struct. All the functions that can take in a `FlatVar` or `Metadata` don't require accessing the data field.

PR #317, opened 2025-08-19
Add Var.has_start_date and Var.start_date

### LINKED ISSUE #316: Add start date function
It would be good to add a start date function that fetch the start date from the attributes if it exists and convert it to a `Dates.DateTime` (and eventually the correct CFTime in the future).

ph-kev:
closes #316 - This PR adds `Var.has_start_date` and `Var.start_date`. The pattern of `Dates.DateTime(var.attributes["start_date"]` is common and this function is helpful to have in the future when `ClimaAnalysis` supports `CFTime` in the future.


ph-kev:
There was also an issue with rebasing in the previous PRs, so this PR also removes a duplicate `include`.

PR #338, opened 2025-09-26
Fix type instability in unflatten

ph-kev:
There was a type instability in `unflatten`, but the compiler could not figure out the number of dimensions of the array because the size of `ordered_dims` was not passed to the compiler.


ph-kev:
> > Is this considered a breaking change? I'm not sure if `Metadata` is considered an internal implementation detail.
> > If type stability is important here. it might be a good idea to add tests for it.
> 
> I consider this part of the `Metadata` as an internal detail. I don't consider the type stability that important, but I will see if it is easy to add a test for these two functions.

The tests work on 1.11 and fail on 1.9, so I ended up removing them.

PR #348, opened 2025-10-23
Add units for binary operations

### LINKED ISSUE #270: Keep units for binary operations
For binary operations like such as addition and subtraction, it would be helpful to keep the units whenever possible. This may need a better units package such as the one suggested in #209.

ph-kev:
closes #270 - This PR makes all binary operations adds units to the attributes when doing a binary operation. Because of the lack of a good units system, the resulting units is not simplified.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 ClimaAnalysis.jl Release Notes
 ===============================
+main
+-------
+
+## Units for binary operations
+
+With this release, all resulting `OutputVar` from binary operations with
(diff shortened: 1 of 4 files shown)


ph-kev:
> It would be nice to combine units in a more sophisticated manner, but this is already much better! For example, if you have `OutputVar x OutputVar` where both have `kg` as the units, it should result in`kg^2` instead of `kg * kg`.

I think if we switch to using `UDUnits.jl`, it might be possible to combine units in a more sophisticated manner. Another solution would be to write a parser.

PR #352, opened 2025-10-30
Add Var.transform_dates

### LINKED ISSUE #351: Make `_shift_by` public
We should export it as well

ph-kev:
closes #351 - This function adds `Var.transform_times` which allows the user to pass a generic function that transform the dates of the `OutputVar`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,19 @@ main
 With this release, all resulting `OutputVar` from binary operations with
 `OutputVar`s and real numbers keep their units when appropriate.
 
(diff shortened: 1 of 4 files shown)


ph-kev:
> Would a user ever want to do the same thing as `transform_dates`, but on the times instead of dates (for example if there is no date/start date)?

In that case, I think it is okay for the user to get `var.dims["time"]` and modify it directly themselves. The dates are a special case, since they are annoying to work with, but maybe there exists a better solution for unifying the handling of times and dates.

PR #354, opened 2025-10-31
Add Var.push_dim

### LINKED ISSUE #353: Add function to add singleton dimension
For calibration (see https://github.com/CliMA/ClimaCalibrate.jl/issues/257), you may want to do a time average, but the time dimension is  gone in this case. To help with this, it would be nice to have a function that insert a singleton dimension to a `OutputVar`.

For example,
```
function insert_singleton_dimension(var::OutputVar, dim_name, dim_value)
       # implementation here
end
```

ph-kev:
closes #353 - This PR adds `Var.push_dim` which adds an extra dimension of length 1 to a `OutputVar`.

### DIFF
--- a/docs/src/howdoi.md
+++ b/docs/src/howdoi.md
@@ -426,7 +426,7 @@ ClimaAnalysis.shift_to_previous_day(var) |> ClimaAnalysis.dates
 ```
 
 These functions are helpful with aligning the dates of observational and
-simulational data.
+simulation data.
 
 ## How do I add a singleton dimension to a `OutputVar`?
 


src/Var.jl
```diff
@@ -2649,6 +2650,38 @@ function Base.cat(vars::OutputVar...; dims::String)
...
+"""
+    push_dim(var::OutputVar, dim_name, dim_value, dim_attrib_pairs...)
+
+Push a singleton dimension to the end of the dimensions in `var`.
+
+If you want to change the order of the dimensions, then call `permutedims` after
+this function.
+"""
+function push_dim(var::OutputVar, dim_name, dim_value, dim_attrib_pairs...;)
+    dims = deepcopy(var.dims)
```
ph-kev: This doesn't work properly if `dim_value` is a `Dates.DateTime`. I am not too sure how crucial that functionality is.

PR #362, opened 2026-01-16
Add support for filenames produced from diagnostics in pressure coordinates

### LINKED ISSUE #360: Add support for reading diagnostics with pressure coordinates
Once this issue (https://github.com/CliMA/ClimaDiagnostics.jl/issues/136) is done, there should be support for diagnostics with pressure coordinates. The problem is that the file name will not be matched by `match_nc_filename` due to the different file name conventions.

https://github.com/CliMA/ClimaAnalysis.jl/blob/b1ab4c14498c1787877d4bf1396ad56c68def0b3/src/Utils.jl#L39

This is, because there is most likely going to be support for writing diagnostics in both z and pressure coordinates at the same time.

ph-kev:
Depends on https://github.com/CliMA/ClimaDiagnostics.jl/issues/136

closes #360

This PR adds support for filenames of the form `shortname_(period)_reduction_(coords).nc`. This functionality is needed when the issue in ClimaDiagnostics is done, as this will add support for diagnostics in pressure coordinates.

TODO

- [x] Add file with the different filename and test with SimDir

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@ ClimaAnalysis.jl Release Notes
 main
 -------
 
+v0.5.21
+-------
+
 ## Units for binary operations
 
 With this release, all resulting `OutputVar` from binary operations with
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaAnalysis"
 uuid = "29b5916a-a76c-4e73-9657-3c8fd22e65e6"
 authors = ["Gabriele Bozzola <gbozzola@caltech.edu>", "Kevin Phan <kphan2@caltech.edu>"]
-version = "0.5.20"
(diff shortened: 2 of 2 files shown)


src/Sim.jl
```diff
@@ -120,6 +134,34 @@ function available_periods(
...
 
+"""
+    available_coords(
+        simdir::SimDir;
+        short_name::String,
+        reduction::String,
+        period::Union{String, Nothing}
+    )
+
+Return the coordinates associated to the given variable, reduction, and period.
```
tapios:
```suggestion
Return the coordinates associated with the given variable, reduction, and period.
```

tapios (approved):
LGTM

imreddyTeja (approved):
Changes look good to me. I left one comment about naming.

PR #372, opened 2026-02-09
Replace Julia v1.9 with 1.12 in CI

ph-kev:
This PR replaces Julia v.1.9 with v1.12 because the long term support release is Julia v1.10.10.


ph-kev:
This was done in https://github.com/CliMA/ClimaAnalysis.jl/pull/413

PR #385, opened 2026-03-11
Add support for multi-file concatenation for `NCCatalog`

### LINKED ISSUE #322: Add functionality for NCCatalog to stitch together multiple datasets together
`NCCatalog` only supports loading a single variable from a single NetCDF file, and cannot aggregate multiple dataset along the time dimension.

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/322 - This PR adds support for multi-file concatenation for `NCCatalog` by reusing the same functionality in `read_var`.

### DIFF
--- a/src/Catalog.jl
+++ b/src/Catalog.jl
@@ -4,7 +4,7 @@ import NCDatasets
 
 import Base: get
 import ..Sim: available_vars
-import ..Var: OutputVar
+import ..Var: OutputVar, _get_time_name_from_mfds
 
 export NCCatalog, add_file!, available_vars
 
@@ -12,14 +12,14 @@ export NCCatalog, add_file!, available_vars
     Organize external NetCDF files to easily initialize `OutputVar`s.
 """
 struct NCCatalog
-    "Dictionary that maps aliases to tuples of file path and variable name in the NetCDF
+    "Dictionary that maps aliases to tuples of filepaths and variable name in the NetCDF
     file. The identifier can either be the variable name or an alias for the variable name
     in the file."
-    identifier_to_path_varname::Dict{String, Tuple{String, String}}
+    identifier_to_path_varname::Dict{String, Tuple{Vector{String}, String}}
 
     "Dictionary that maps variable names to filepaths. This is a fallback if a variable is
     requested and cannot be found in `identifier_to_path_varname`."
-    varname_to_filepath::Dict{String, String}
+    varname_to_filepath::Dict{String, Vector{String}}
 end
 
 """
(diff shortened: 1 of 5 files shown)


imreddyTeja:
What does mfds stand for? Multi-file data struct?

ph-kev:
> What does mfds stand for? Multi-file data struct?

Multi file dataset

src/Catalog.jl
```diff
@@ -58,30 +58,92 @@ function add_file!(catalog::NCCatalog, filepath, short_names...)
...
+        ArgumentError(
+            "There is at least one file in $filepaths that is not a NetCDF file, because the file path does not end with .nc",
+        ),
+    )
+
+    maybe_short_name_pairs = short_names
+    if isempty(maybe_short_name_pairs)
+        short_name = nothing
+    else
+        maybe_short_name_pair = first(short_names)
```
imreddyTeja: I know this is not a new addition, but I'm a bit confused on what this is doing

imreddyTeja (approved):
Would this be a breaking change because ` NCCatalog` is an exported struct?

src/Catalog.jl
ph-kev: ~~I need to get a short name to figure out what the time dimension is. There is probably a better way of doing this now that I think about it.~~

ph-kev:
> Would this be a breaking change because ` NCCatalog` is an exported struct?

I don't think this will be a breaking change? There are changes to the internals of the struct, but none of our code make use of the internals and they shouldn't anyway.

src/Catalog.jl
ph-kev: I used the short name since I needed to check if the `start_date` attribute is consistent between all the datasets. I think a better way of doing this is to scan all the variable names and check that all of them has a start date or don't has a start date.

src/Var.jl
```diff
@@ -490,6 +440,80 @@ function read_var(paths::Vector{String}; short_name = nothing)
     end
 end
 
+"""
+    _get_time_name_from_mfds(paths)
+
+Get the name of the time dimension in the NetCDF files in `paths`.
+
+If a `short_name` is specified, then the short name will be used to access the
```
imreddyTeja, line 448: Is this saying that it will use the time dim of the variable with short name `short_name`?

imreddyTeja (approved):
One nit about some wording

src/Var.jl
ph-kev: No. There is a list of names for the time dimensions and we use that to identify what is the time dimension for the NetCDF file. The short name is only used to check for the start date attribute which is part of the variable and not part of the time dimension.

PR #389, opened 2026-04-07
Clean up typos and missing type annotations

### LINKED ISSUE #365: Missing keyword argument for docs of `variance_*` functions
The missing keyword argument is `corrected`. The documentation said that you can pass it in, but the function signature is missing it.

### LINKED ISSUE #374: Keep units for unary minus for OutputVar
When you write `-var` where `var` is a `OutputVar`, then units are removed. For most unary functions, this make sense, but units should be kept for this case.

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/374, closes https://github.com/CliMA/ClimaAnalysis.jl/issues/365

This PR keeps units for unary minus for `OutputVar` (flipping the sign of the data is common for postprocessing), adding missing type annotations, and fixing typos in the docs.

### DIFF
--- a/src/outvar_operators.jl
+++ b/src/outvar_operators.jl
@@ -3,7 +3,11 @@
 
 Add methods to overload the given binary `op`erator for `OutputVars` and `Real`s.
 
-Attributes that are not `short_name`, `long_name`, are discarded in the process.
+Attributes that are not `short_name`, `long_name`, `start_date`, and `units` are discarded
+in the process.
+
+The `start_date` and `units` are only kept if the start date and units are the same
+respectively.
 """
 macro overload_binary_op(op)
     quote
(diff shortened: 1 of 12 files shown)


src/outvar_operators.jl
```diff
@@ -167,9 +171,13 @@ end
 
 Generate a method to overload the unary operator `op` for `OutputVar`.
 
-Handles attributes `short_name`, `long_name`: Prepends the operator name, e.g., "log(Temperature)".
+Handles the attributes `short_name`, `long_name`, `start_date`, and `units` and prepends the
+operator name, e.g., "log(Temperature)".
+
+The attribute `units` is only kept for unary minus.
```
imreddyTeja, line 177: Should this also be the case for `abs`?

imreddyTeja (approved):

src/outvar_operators.jl
ph-kev: `abs` isn't defined for `OutputVar`s.

PR #390, opened 2026-04-07
Add getters for dimension names

ph-kev:
This PR adds getters for dimension names. Users would use `keys(var.dims)` to get the dimension names, so users can now do `Var.dim_names(var)` instead.

This PR used to include a getter for the dimension values, but that function was removed. That functionality was accomplished by `longitudes`, `latitudes`, etc.


nefrathenrici (approved):
Thank you!

PR #391, opened 2026-04-21
Add check for start date when resampling

### LINKED ISSUE #315: Add check for start date in `resampled_as`
There is no check for start date when using `resampled_as`. It would be reasonable to check that the start date are both the same or both are not present, so that `resampled_as` is correct.

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/315

In the current version of ClimaAnalysis, there is no check for the start date. The user was expected to check the start date themselves and this behavior was not documented. This could leads to silent errors in the postprocessing. This PR fixes this by checking the start date of the source and destination `OutputVar`s and handling the different cases of the `OutputVar`s having a start date or not.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -1614,6 +1614,10 @@ longitude dimension because `conventional_dim_name` maps `lon`, `long`, and `lon
 !!! compat "`dim_names` keyword argument"
     The keyword argument `dim_names` is introduced in ClimaAnalysis v0.5.14.
 
+!!! note "Reference date"
+    If `src_var` and `dest_var` do not have the same reference or start date,
+    then the reference date of `src_var` is set to the reference date of
+    `dest_var` before resampling.
 """
 function resampled_as(
     src_var::OutputVar,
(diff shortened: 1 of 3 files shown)


src/Var.jl
```diff
@@ -1754,6 +1764,52 @@ function resampled_as(src_var::OutputVar; kwargs...)
...
 end
 
+"""
+    _ensure_consistent_reference_date(
+        src_var::OutputVar,
+        dest_var::OutputVar,
+    )
+
+Ensure that `src_var` and `dest_var` have the same reference date or no reference date. If
+either `OutputVar`s have a start date and the other one doesn't, then throw an error.
```
imreddyTeja: nit: maybe mention that `src_var` is returned if ref dates are consistent

imreddyTeja (approved):

src/Var.jl
ph-kev: I can add better docs for this function.

PR #396, opened 2026-05-15
Implement subset of split-apply-combine pattern

### LINKED ISSUE #279: Better support for group by and reduce by
It would be nice to have better support for grouping along a single dimension and applying a reduction for along that dimension. 

For example, in calibration, one application is grouping by seasons across time, applying a time average for each season, grouping everything by seasons, and applying a variance operation.

As of now, this is difficult to do because the need to handle attributes and dimension attributes. Furthermore, it is not obvious what the values should be along the reduced dimension.

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/279

This PR implements a subset of the split-apply-combine pattern which are supporting reductions, single group split, and seasonal splits. This allows the users to be more expressive with the kind of reductions that they can do. The implementation of the split-apply-combine pattern also keep the dimension that is reduced over which can be helpful for calibration (e.g. certain calibration functionality such as `ObservationRecipe` requires a time dimension).

Here's a list of concerns that I have for this PR:

1. I wasn't sure if this should be in a module or not. I decided against it since putting the functionality in a module isn't a breaking change as long as I export the fucntion.
2. I couldn't get the docs for the struct behaving as a function working properly. I think the example is sufficient.
3. There is no support for reductions that require knowledge of the dimension values. This can be done in a later PR if the need arises and can be implemented as another `AbstractApplyOperation`.
4. Error handling happens at materialization instead of at the construction of the `SplitApplyVar`. This can be changed later without it being a breaking change.
5. The element of the new dimension is always the first element of each group. I don't think there need to be support for anything else, since I can't think of another use case for this.
6. The attributes and dimension attributes are not updated. I wasn't too sure about this.

In the future, for the split operations, there would be support for arbitrary splits that the user can supply. Additionally, the apply operations can also include transformations (which keep the size of the array the same) and filter (requires adding a new dimension). An example of a transformation is doing a z-score transformation each group (e.g. seasons) and an example of a filter is getting the top k elements of each group. Both use cases are not needed right now, so they are not included in this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,26 @@ ClimaAnalysis.jl Release Notes
 main
 -------
 
+## Split-apply-combine
+
+This release features an initial implemention of the split-combine-apply
+pattern. For example, to do a time average, you can now do
+
+```julia
+import ClimaAnalysis
+import Statistics: mean
+
+reduced_var =
+    var |>
+    ClimaAnalysis.GroupAll("time") |>
+    ClimaAnalysis.Reduce(mean) |>
+    ClimaAnalysis.combine
+```
+
+The resulting `OutputVar` has a time dimension containing the first time value
+of `var`. For more information, see the section "Split-Apply-Combine" in the
+documentation.
+
 ## Bug fixes
 
 - Enforce uniqueness when splitting dates by seasons across time.
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ main
 
 ## Bug fixes
 
+- Enforce uniqueness when splitting dates by seasons across time.
 - Exclude NaNStatistics v0.6.57 in compat, because of correctness issues and
   possibility of a segfault. See this
   [issue](https://github.com/brenhinkeller/NaNStatistics.jl/issues/66) in
--- a/docs/src/split_apply_combine.md
+++ b/docs/src/split_apply_combine.md
@@ -1,0 +1,103 @@
+# Split Apply Combine
+
+!!! note "Experimental"
+    This is an initial implementation of the split-apply-combine pattern that
+    supports a limited set of features. As a result, the public API will not be
+    covered under semantic versioning and is subjected to changes, but we will
+    try not to change it.
+
+    If you need a feature that is not yet implemented, please let us know!
+
+The paper ["The Split-Apply-Combine Strategy for Data Analysis"](https://www.jstatsoft.org/article/view/v040i01),
+written by Hadley Wickham, illustrates that most data processing follows the
+pattern of:
+- splitting data into groups,
+- applying a reduction, transformation, or filter on each group,
+- combining the results together via concatenation.
+
+ClimaAnalysis currently implements splitting operations via a single group via
+[`GroupAll`](@ref) or seasonal splits via [`SplitSeason`](@ref) and reductions
+via [`Reduce`](@ref).
+
+## Tutorial
+
+The split-apply-combine pattern is only defined for `OutputVar`s. Assuming you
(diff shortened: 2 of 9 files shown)


docs/src/split_apply_combine.md
```diff
@@ -0,0 +1,102 @@
...
+import Statistics: mean
+
+time_averaged_var =
+    var |>
+    ClimaAnalysis.GroupAll("time") |>
+    ClimaAnalysis.Reduce(mean) |>
+    ClimaAnalysis.combine
+```
+
+!!! note "What is the difference between this example and [`average_time`](@ref)?"
```
imreddyTeja: This ref is not rendering correctly

docs/src/split_apply_combine.md
```diff
@@ -0,0 +1,102 @@
...
+    There are no modifications to the attributes in this example and
+    `average_time` modifies the attributes to reflect that the operation
+    happens. Also, `average_time` squeezes the time dimension and removes it
+    while the time dimension is kept when using the split-apply-combine pattern.
+
+First, an [`AbstractSplitOperation`](@ref) is called on a `OutputVar` which
+produces [`SplitApplyVar`](@ref). In this example, `var` is piped to
```
imreddyTeja:
```suggestion
produces a [`SplitApplyVar`](@ref). In this example, `var` is piped to
```

src/split_apply_combine.jl
```diff
@@ -0,0 +1,255 @@
...
+    # number of groups
+    # We will always get the first element of the dimension
+    ret_dim_indices = (first(dim_indices) for dim_indices in dim_indices_groups)
+
+    # New dimension to return
+    dim_name_in_var = var.index2dim[dim_idx]
+    dim = [var.dims[dim_name_in_var][i] for i in ret_dim_indices]
+
+    # Make new dimensions for OutputVar
+    ret_dims = deepcopy(var.dims)
```
imreddyTeja, line 250: I'm a bit confused on what this does.

imreddyTeja (approved):
This looks good. I like this design pattern, but I don't really understand what the `combine` part does.

test/test_split_apply_combine.jl
nefrathenrici, line 1: Nitpick but this file could be renamed to `test_split_apply_combine.jl`

NEWS.md
```diff
@@ -3,6 +3,29 @@ ClimaAnalysis.jl Release Notes
...
+    ClimaAnalysis.Reduce(mean) |>
+    ClimaAnalysis.combine
+```
+
+The resulting `OutputVar` has a time dimension containing the first time value
+of `var`. For more information, see the section "Split-Apply-Combine" in the
+documentation.
+
+## Bug fixes
+- Enfore uniqueness when splitting dates by seasons across time.
```
nefrathenrici:
```suggestion
- Enforce uniqueness when splitting dates by seasons across time.
```

src/split_apply_combine.jl
```diff
@@ -0,0 +1,255 @@
...
+end
+
+"""
+    combine(split_apply_var::SplitApplyVar)
+
+Materialize the `OutputVar` with the split and apply operations.
+
+This splits the `OutputVar` into groups via an `AbstractSplitOperation`, applies
+an `AbstractApplyOperation` to each group, and combine the results together. The
+split dimension is is kept. The coordinate values of the split dimension in the
```
nefrathenrici:
```suggestion
split dimension is kept. The coordinate values of the split dimension in the
```

nefrathenrici (approved):
Looks good! Just nits

PR #403, opened 2026-06-08
Add Base.dropdims for OutputVars

### LINKED ISSUE #395: Add `dropdims`
It would be nice if there is a function similar to `Base.dropdims` for `OutputVar`s.

ph-kev:
This PR adds `Base.dropdims` for `OutputVar`s. This is meant to support workflows using the split-apply-combine pattern, since there is no easy way of getting rid of the singleton dimensions.

closes https://github.com/CliMA/ClimaAnalysis.jl/issues/395

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -41,6 +41,7 @@ Var.dim_names
 Var.units
 Var.has_units
 Var.remake
+Base.dropdims
 Var.slice
 Var.average_lat
 Var.weighted_average_lat
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -48,6 +48,7 @@ export OutputVar,
     dim_units,
     range_dim,
     ndims,
+    dropdims,
     reordered_as,
     resampled_as,
     has_units,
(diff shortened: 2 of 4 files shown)


docs/src/api.md
```diff
@@ -41,6 +41,7 @@ Var.dim_names
 Var.units
 Var.has_units
 Var.remake
+Base.dropdims(var::OutputVar; dims)
```
imreddyTeja: just wondering, how do you decide when to add types to the api docs

src/Var.jl
```diff
@@ -1191,6 +1192,37 @@ function Base.ndims(var::HasDimAndAttribs)
     return length(var.dims)
 end
 
+"""
+    dropdims(var::OutputVar; dims)
+
+Return a `OutputVar` with the same data as `var`, but with the dimensions
+specified by `dims`, an iterable of dimension names, removed. The size of each
```
imreddyTeja, line 1199: very minor nit, and this is probably mentioned somewhere else, but should dims be an iterable of strings or symbols?

src/Var.jl
```diff
@@ -1191,6 +1192,37 @@ function Base.ndims(var::HasDimAndAttribs)
...
+The result shares the same underlying data as `var` which means modifying values of
+`var.data` also results in the same changes to the data of the resulting `OutputVar`.
+"""
+function Base.dropdims(var::OutputVar; dims)
+    dim_names = collect(
+        find_corresponding_dim_name_in_var(dim_name, var) for dim_name in dims
+    )
+
+    dim_indices = Tuple(var.dim2index[dim_name] for dim_name in dim_names)
+    ret_data = dropdims(var.data; dims = dim_indices)
```
imreddyTeja, line 1222: is the error thrown by Base.dropdims when there is a non singleton dimension informative enough?

imreddyTeja (approved):

src/Var.jl
ph-kev: I don't mention it, but I don't think this detail is important to have. Other functions like `permutedims` don't mention it.

docs/src/api.md
ph-kev: I specified the type signature for `Base` methods, since I remembered in the past, it would also give other methods that I didn't want. I switched it to `Base.dropdims` which work.

src/Var.jl
ph-kev: I improved the error messages

PR #407, opened 2026-06-22
Add equality and hashing for `OutputVar`, `FlatVar`, and `Metadata`

### LINKED ISSUE #406: Define equality for OutputVar, FlatVar, and Metadata
The function `Base.isequal` is not defined for `OutputVar`, `FlatVar`, and `Metadata`. As a result, something like

```julia
var2 = deepcopy(var1)
isequal(var1, var2) # return False
```

would return `False` even though the contents of `var1` and `var2` are the same. This might requires defining some combination of `Base.:(==)`,`Base.hash`, and `Base.isequal`. Since `Base.isequal` calls `Base.:(==)`, it might be sufficient to just define the first two functions.

ph-kev:
closes #406 - This PR adds equality and hashing for `OutputVar`, `FlatVar`, and `Metadata`.

TODO

- [x] Add docs for equality and hashing functions added

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -2998,6 +2998,56 @@ function Base.identity(var::OutputVar)
     return var
 end
 
+"""
+    Base.:(==)(var1::OutputVar, var2::OutputVar)
+
+Return whether `var1` and `var2` are equal, using `==` on each field.
+
+This is similar to `isequal`, except `==` treats floating-point `NaN` values as not equal to
+each other, `-0.0` as equal to `0.0`, and `missing` as not equal to `missing`.
+"""
+function Base.:(==)(var1::OutputVar, var2::OutputVar)
+    # Use & instead of &&, since we want to propagate missing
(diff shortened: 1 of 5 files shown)


src/Var.jl
```diff
@@ -2998,6 +2998,50 @@ function Base.identity(var::OutputVar)
...
+           (var1.dim_attributes == var2.dim_attributes) &
+           (var1.data == var2.data) &
+           (var1.dim2index == var2.dim2index) &
+           (var1.index2dim == var2.index2dim)
+end
+
+"""
+    Base.isequal(var1::OutputVar, var2::OutputVar)
+
+Return whether `var1` and `var2` are equal, using `isequal` on each field.
```
imreddyTeja, line 3022: I think it would be good to add a quick note to both methods explaining the relevant differences in behaviors for outputvars

imreddyTeja (approved):
left one nit about the docstrings for `==` vs `isequal`

PR #413, opened 2026-07-10
Update to JuliaFormatter v2.10.1

ph-kev:
This PR updates JuliaFormatter to v2.10.1 and pin it. This is recommended by https://juliaeditorsupport.github.io/JuliaFormatter.jl/stable/status/#Pinning-the-JuliaFormatter-version.


nefrathenrici (approved):

Project.toml
imreddyTeja, line 1: Are the compat changes here because the downgrade test would fail?

imreddyTeja (approved):

Project.toml
ph-kev: Yes. I think both Aqua, ExplicitImports, and NCDatasets rely on internals that Julia 1.12 remove. I think NaNStatistics rely on a package that uses internals that Julia 1.12 remove. I think Unitful set compat bounds for Julia 1.9 and Julia 1.10.

PR #415, opened 2026-07-16
Remove allocations

ph-kev:
I asked Claude to find any easy fixes to allocations. The changes are not materializing vectors when constructing dictionaries, using more idiomatic functions like `issetequal`, and removing unnecessary copying.

### DIFF
--- a/src/flat.jl
+++ b/src/flat.jl
@@ -254,10 +254,10 @@ function unflatten(metadata::Metadata, data::AbstractVector)
     unflattened_data = reshape(flat_data, reshape_dims...)
 
     # Permute dimensions of data
-    dim2index = Dict([
+    dim2index = Dict(
         dim_name => index for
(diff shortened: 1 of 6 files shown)


src/flat.jl
```diff
@@ -267,7 +267,7 @@ function unflatten(metadata::Metadata, data::AbstractVector)
         deepcopy(metadata.attributes),
         deepcopy(metadata.dims),
         deepcopy(metadata.dim_attributes),
-        copy(data),
```
imreddyTeja, line 270: Is this behavior changing?

imreddyTeja (approved):
I didn't know `any` could be used like that. Left a nit about potential behavior change that might need documentation

src/flat.jl
ph-kev: This should be fine since the line before this is `data = permutedims(unflattened_data, perm)` which always allocates a new array.

PR #417, opened 2026-07-20
Remove hard coded examples in documentation

ph-kev:
This PR removes some of the hard coded examples in the documentation and replaced them with Documenter.jl's `@example` or `@repl` to run the code when generating the documentation. This helps keep the documentation up to date in case if the function signature change or the behavior of the functions change. However, this does marginally increase the time it takes for the documentation to be created.

I asked Claude to make these changes.

TODO

- [x] Improve visualization documentation


imreddyTeja (approved):
lgtm

PR #418, opened 2026-08-03
Add interpolation module

### LINKED ISSUE #182: Var.resampled_as too much allocations

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/182 - This PR adds the module Interpolation which implements its own regridder for linear interpolation on CPU. This speeds up `resampled_as` as shown in the benchmark below. This is also needed to implement `NaN`-aware interpolation which will build on top of this PR. 

```
# ClimaCoupler leaderboard plots comparsion using @time
# Current release of ClimaAnalysis
# First call
633.497110 seconds (4.78 G allocations: 350.271 GiB, 11.06% gc time, 25.36% compilation time: 11% of which was recompilation)

# Second call
453.559515 seconds (4.48 G allocations: 336.167 GiB, 9.66% gc time, 0.05% compilation time: 37% of which was recompilation)

# This branch
# First call
226.008800 seconds (326.68 M allocations: 142.870 GiB, 11.99% gc time, 52.08% compilation time: 15% of which was recompilation)

# Second call
95.502202 seconds (47.26 M allocations: 129.401 GiB, 16.47% gc time, 0.24% compilation time: 37% of which was recompilation)
```

### DIFF
--- a/src/Interpolation.jl
+++ b/src/Interpolation.jl
@@ -146,16 +146,16 @@ function Regridder(
 
     for (dim_idx, coords) in enumerate(src_grid)
         length(coords) >= 2 || error(
-            "The $(dim_idx)th dimension has less than two coordinates, which is not supported",
+            "Dimension $dim_idx has less than two coordinates, which is not supported",
         )
 
         # NaN and missing values in coordinates are not supported
         any(x -> ismissing(x) || isnan(x), coords) && error(
-            "Missing or NaN values for the $(dim_idx)th dimension is not supported",
+            "Missing or NaN values for dimension $dim_idx is not supported",
         )
 
         issorted(coords, lt = <=) ||
-            error("The $(dim_idx)th dimension  is not strictly increasing")
+            error("Dimension $dim_idx is not strictly increasing")
     end
 
     for (dim_idx, (coords, extp)) in enumerate(zip(src_grid, extrapolation))
--- a/src/Interpolation.jl
+++ b/src/Interpolation.jl
@@ -1,0 +1,582 @@
+module Interpolation
+
+import OrderedCollections: OrderedDict
+
+"""
+    Node
+
+A staggering where the coordinates are the edges of the cells, so the first
+and last coordinates are the boundaries of the domain.
+"""
+struct Node end
+
+"""
+    Center
+
+A staggering where the coordinates are the centers of the cells, so the domain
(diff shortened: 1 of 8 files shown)


src/Var.jl
```diff
@@ -118,104 +118,65 @@ include("flat.jl")
...
+    # one coordinate
     if isempty(dims) || any(d -> ndims(d) != 1 || length(d) == 1, values(dims))
         return nothing
     end
 
-    # Dimensions are all 1D, check that the knots are in increasing order (as required by
-    # Interpolations.jl)
+    # Check that the coordinates are in increasing order
     for (dim_name, dim_array) in dims
         if !issorted(dim_array)
```
nefrathenrici: This isn't changed but this won't warn for dimensions that are not strictly increasing, such as `[1.0, 1.0, 2.0]`. A more consistent check would be `issorted(dim_array, lt = <=)`

src/Interpolation.jl
```diff
@@ -0,0 +1,582 @@
...
+
+"""
+    _interior_stencil(x::Number, coords::AbstractVector)
+
+Return the stencil `(index1, index2, w)` for `x` and `coords` when `x` is in
+the domain of `coords`.
+"""
+@inline function _interior_stencil(x::Number, coords::AbstractVector)
+    i = searchsortedfirst(coords, x)
+    first_index = firstindex(coords)
```
nefrathenrici, line 384: I don't think it's technically needed, but you could add a fast path for exact matches:
```suggestion
    first_index = firstindex(coords)
    coords[i] == x && return (i, i, one(x / x))
```

nefrathenrici (approved):
This looks good to me! One non-blocking idea: the stencil precomputation is clearly worth it, but keeping the cache as a field on Regridder is only useful if a regridder gets reused, and nothing here reuses one. Making it a local in `regrid!` would also make Regridder immutable.

src/Interpolation.jl
ph-kev: I don't think this will have a big impact on performance. It does make the NaN case better though, but I think in another PR, we can implement NaN aware regridding instead of trying to improve it here.

PR #425, opened 2026-08-05
Exclude NaNs from normalization in error metrics

### LINKED ISSUE #424: Improve masking for `bias`
When dealing with masked data, it can be improved by using the `sim - obs` and generating the normalization from that. Right now, it always assume the data is all ones. However,  this can be improved by using `NaN`s when there is a corresponding `NaN`s in `sim - obs`.

ph-kev:
fixes #424 - This bug was found in https://github.com/CliMA/ClimaLand.jl/pull/1836. For `OutputVar`s with `NaN`s in the data, the error metrics could have been computed wrong if a mask function is not given. This is because the normalization is computed using the mask function. This PR fixes that issue by computing the normalization from the `NaN`s in `sim - obs`.

I asked a LLM to implement these changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,6 +58,11 @@ Makie.save("surface_plot.png", fig)
 - Masks generated from `generate_lonlat_mask` are now recognized as masking
   functions when plotting (e.g. `plot_bias_on_globe!`). Previously, they were
   not recognized, so plotting proceeded without applying the mask.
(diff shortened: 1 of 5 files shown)


nefrathenrici (approved):
This looks good to me, I think we could follow up on this PR to handle NaNs more uniformly.

PR #426, opened 2026-08-05
Add abs as a unary operation for OutputVar

### LINKED ISSUE #423: Add `abs` for `OutputVar`
You should be able to do something like `abs(var)` where `var` is a `OutputVar`.

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/423 - This PR adds `abs` as another unary function for `OutputVar`.


nefrathenrici (approved):
Thanks!

PR #429, opened 2026-08-07
Add NaN-aware resampling

### LINKED ISSUE #138: Alternatives to resampling with land observational data
Resampling don't work as well as it should be for observational data. The first plot is of observational data for `gpp`. The second plot is after resampling onto simulation data. Because there are `NaN`s in the data, resampling don't work quite as well as the `NaN`s poison the values near the boarder. One potential solution is to resample on the union of the two grids so that no information is lost when going to the new grid.

gpp
gpp_resampled_as

### LINKED ISSUE #198: Add mask-aware interpolation
Similar to this [issue](https://github.com/CliMA/ClimaLand.jl/issues/987), it would be nice to have mask-aware interpolation for masking `OutputVar`s and plotting.

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/138 and closes https://github.com/CliMA/ClimaAnalysis.jl/issues/198 - This PR adds NaN aware regridding with the `nan_threshold` keyword argument. As `nan_threshold` increases, the number of `NaN`s decrease. 

This is done by summing the weights that the `NaN`s contributed. If the sum is greater than the `NaN` threshold, then the resulting value is a `NaN`. Otherwise, it is the weighted average of the non-`NaN` values.

A LLM implemented this and I went through and verified it.

TODO

1. Think about how `resampled_as` should be handled for functions that compute error metrics

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,23 @@ ClimaAnalysis.jl Release Notes
 main
 -------
 
+## NaN-aware resampling
+
+The `resampled_as` functions now accept the keyword argument `nan_threshold`
+to control how `NaN`s in the source data are handled.
+
+```julia
+resampled_var = ClimaAnalysis.resampled_as(src_var, dest_var, nan_threshold = 0.5)
+```
+
+By default, a resampled value is `NaN` whenever any of the values it
+interpolates from is `NaN`, so a few `NaN`s in the source data can spread to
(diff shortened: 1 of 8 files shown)


imreddyTeja (approved):
reviewing this made me appreciate a defined private/public interface

PR #431, opened 2026-08-19
Add Var.propagate_nans

ph-kev:
This PR adds `Var.propagate_nans` which let the users propagate `NaN` across the slices of a `OutputVar`. This functionality is used in calibration because the observational data has `NaN`s at different non-temporal coordinates across the time slices of the `OutputVar`. To build a matrix of samples, we need to either do imputation or propagate the `NaN`s over the samples. The latter solution is implemented.

### DIFF
--- a/src/Var.jl
+++ b/src/Var.jl
@@ -79,6 +79,8 @@ export OutputVar,
     set_reference_date!,
     replace,
     replace!,
+    propagate_nans,
+    propagate_nans!,
     reverse_dim,
     reverse_dim!,
     remake,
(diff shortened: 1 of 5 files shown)


src/Var.jl
```diff
@@ -2751,6 +2753,52 @@ function Base.replace!(
...
+    FT = eltype(var.data)
+    nan_mask = map(
+        contains_nan -> contains_nan ? FT(NaN) : one(FT),
+        any(isnan, var.data, dims = dim_indices),
+    )
+    var.data .*= nan_mask
+    return nothing
+end
+
+
```
nefrathenrici:
```suggestion
```
nit

src/Var.jl
```diff
@@ -2751,6 +2753,52 @@ function Base.replace!(
...
+function propagate_nans!(var::OutputVar; dims = ("time",))
+    dims isa AbstractString && (dims = (dims,))
+    dim_names = collect(
+        find_corresponding_dim_name_in_var(dim_name, var) for dim_name in dims
+    )
+    allunique(dim_names) || error("Dimensions ($dim_names) must be unique")
+
+    dim_indices = Tuple(var.dim2index[dim_name] for dim_name in dim_names)
+
+    FT = eltype(var.data)
```
nefrathenrici: Isn't it possible for this to be Union{Missing,Float64}, or are we expected to remove missing values before calling `propagate_nans!` ?

nefrathenrici (approved):
Thanks! Since this is new public API, could you add a NEWS.md entry as well?

src/Var.jl
ph-kev: I fixed this and I expect the users to remove the missing values themselves. One strange thing is that `missing * NaN == missing`, so `propagate_nans` doesn't propagate NaNs if there is a `missing` value. I am not sure what behavior we want here. The current behavior is to preserve the `missing` values.

PR #436, opened 2026-09-10
Add functions to check for missing units

### LINKED ISSUE #433: Add function to check for missing units
It would be nice to have a function that check for missing units of data and dimensions (e.g. `is_missing_units(var::OutputVar)` and `is_missing_dim_units(var::OutputVar, dim)`).

ph-kev:
closes https://github.com/CliMA/ClimaAnalysis.jl/issues/433 - This PR adds functions to check for the units and dimension units of `OutputVar` and `FlatVar`.

I used a LLM to generate the code.


nefrathenrici (approved):
Thanks! It looks like `is_missing_units` and `has_units` disagree - what is `has_units` used for and can it be removed?

ph-kev:
> Thanks! It looks like `is_missing_units` and `has_units` disagree - what is `has_units` used for and can it be removed?

We can't remove it since it is part of the public API. `has_units` just check if the `units` attribute exist which is different from checking that units exist.

